PR #2, opened 2020-06-24
Release 1.3.0

angelbarrera92:
This PR aims to release Fury v1.3.0 adding support to Kubernetes 1.16, 1.17 and 1.18.
Is not a big release in terms of features, but is a step forward to have an updated supported distribution (and certified).

I will work on pending stuff after merging the pr (technical release).

We can promote the new version once:

- docs site is updated
- CNCF verifies our distribution
- having a blog post

Thanks!

### DIFF
--- a/.rules/.markdown-lint.yml
+++ b/.rules/.markdown-lint.yml
@@ -1,0 +1,36 @@
+---
+###########################
+###########################
+## Markdown Linter rules ##
+###########################
+###########################
+
+# Linter rules doc:
+# - https://github.com/DavidAnson/markdownlint
+#
+# Note:
+# To comment out a single error:
+#   <!-- markdownlint-disable -->
+#   any violations you want
+#   <!-- markdownlint-restore -->
+#
+
+###############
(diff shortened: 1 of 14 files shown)


.rules/.markdown-lint.yml
```diff
@@ -0,0 +1,36 @@
...
+#
+
+###############
+# Rules by id #
+###############
+MD004: false                  # Unordered list style
+MD007:
+  indent: 2                   # Unordered list indentation
+MD013:
+  line_length: 808            # Line length
```
iknite, line 25: Why we want a 808 line length?

angelbarrera92: I copy pasted the default `.markdown-lint.yml` file from super-linter and i just added the 

```yaml
 MD041: false 
```

Because we have markdown files does not start with a top level header `#` instead `###`. (Release notes markdown files)

If we want to have a small line width, feel free to submit a PR to our "policeman"

PR #6, opened 2021-09-23
Add commonLabels to all KFD resources

nandajavarma:
The labels will be useful to identify each resource as a part of KFD and which release of KFD is deployed. The assumption is that, the `vendor/` directory will be populated by `furyctl` so this installation is managed by `furyctl`. The version of `furyctl` is hard to assess at this point and hence not including that.

### DIFF
--- a/kustomization.yaml
+++ b/kustomization.yaml
@@ -2,6 +2,12 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
+commonLabels:
+  app.kubernetes.io/part-of: "kfd"
+  kfd.sighup.io/version: "1.7.0"
+  kfd.sighup.io/source: "kustomize"
+  kfd.sighup.io/installer: "furyctl"
+
 bases:
   # Networking
   - ./vendor/katalog/networking/calico


lnovara (changes requested):
Honestly, I would change the `app.kubernetes.io/part-of` to something else since there is an high chance that some applications in the cluster will have a `app` label that might cause conflicts with Elasticsearch labels typing.

Also, adding @nutellinoit to have additional feedback.

nutellinoit:
> Honestly, I would change the `app.kubernetes.io/part-of` to something else since there is an high chance that some applications in the cluster will have a `app` label that might cause conflicts with Elasticsearch labels typing.
> 
> Also, adding @nutellinoit to have additional feedback.

There should be no problem with the new filter on fluentbit: https://github.com/sighupio/fury-kubernetes-logging/blob/master/katalog/fluentd/scripts/dedot.lua . Before doing this change i advise to do a manual test that everything is ok.

PS: BTW I have mixed feelings about forcing labels on the distribution level. Each module should know which is its distribution parent version. This way, we can find mixed versions on target clusters easily.

lnovara:
Additionally, we rarely use this repository while setting up clusters because with furyctl we rely on the fury-kubernetes-* repositories. But maybe this is going to change in the future.

SIGHUP-C-3PO:
@lnovara I will fix the `part-of` for sure.

About the labels being here or modules part, the release version of KFD will be sticky to a release of module. And according to the new plan, modules should have a dev/release cycles independent of KFD release. If module is being handled without this repo, ideally it would be good to not have KFD version number, but just the version of the module and of the components. This is just my perspective. I will have more discussions on this before merge.

nandajavarma:
@lnovara I will fix the part-of for sure.

About the labels being here or modules part, the release version of KFD will be sticky to a release of module. And according to the new plan, modules should have a dev/release cycle independent of KFD release. If module is being handled without this repo, ideally it would be good to not have KFD version number, but just the version of the module and of the components. This is just my perspective. I will have more discussions on this before merge.

nandajavarma:
According to @ipedrazas's suggestion regarding this, it is clear that it wouldn't make much sense to have KFD module version hard-coded into modules since it can be confusing and restrict a lifecycle on its own. Closing this in favor of opening PRs to add labels to modules.

PR #10, opened 2022-01-21
Adding PR and issue templates to fury-distro

nandajavarma:


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,107 @@
+name: Bug Report
+description: Report a bug encountered while operating Kubernetes
```
ralgozino, line 2: If this template will be the same for all the distro repos (the distro itself and the modules) maybe we should talk about operating kubernetes? we could just use "the software" maybe?

ralgozino (changes requested):

nandajavarma:
@ralgozino I have fixed the issue you mentioned. More discussion has to be done on this, I think. I have added a sample script here which can be used to sync the files to different modules. I do not have the permissions to test it out properly.

ralgozino:
Hi @nandajavarma ,Ping me on slack and let's see the permissions thing together

PR #30, opened 2022-12-14
Furyctl next integration

omissis:


kfd.yaml
```diff
@@ -0,0 +1,35 @@
...
+  dr: v1.10.1
+  ingress: v1.13.1
+  logging: v3.0.1
+  monitoring: v2.0.0
+  opa: v1.7.3
+  networking: v1.10.0
+kubernetes:
+  eks:
+    version: 1.23
+    installer: v1.9.1
```
omissis, line 18: update to v1.10.0

kfd.yaml
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+version: v1.25.1
```
nutellinoit, line 5: This version should be v1.25.2

source/manifests/networking/kustomization.yaml.tpl
```diff
@@ -0,0 +1,7 @@
+---
```
nutellinoit, line 1: This file is a leftover, I think should be removed

nutellinoit (changes requested):
overall LGTM, there are some missing points to be done:

- Write release note for v1.25.2 under `docs/releases`
- Write upgrade guide under `docs/upgrades` (the upgrade guide will only say that nothing changes module wise, but we added the alpha support for Furyctl)
- Add Furyfile.yml and kustomization.yaml on `releases/v1.25.2` (and maybe the new kfd.yaml)
- Update main README.md with the v1.25.2 version
- Update the compatibility matrix

nutellinoit (approved):
LGTM!

PR #34, opened 2022-12-19
feature/schema refinements

omissis:
- feat: refine schema validation, introducing conditionals
- feat: introduce schema conditional statements and a bunch of testcases to cover them


nutellinoit (approved):
LGTM, we need to remember to change 1.23 to 1.24 on `kfd.yaml`

PR #39, opened 2023-01-12
Docs: new architecture image

ralgozino:
Use the same image for explaining KFD's modular architecture that we use in our marketing material so we need to maintain just one image.

This change has been approved by Marketing.


nutellinoit (approved):
Very LGTM 🎉

PR #44, opened 2023-01-17
fix: fix template (identation for pomerium/dex template, add baseDomain on auth, fix opensearch-dashboards key)

nutellinoit:
As per title


ralgozino (approved):
LGTM

PR #45, opened 2023-01-17
feat: added baseDomain to spec.distribution.modules.auth

Al-Pragliola:
Changelist:

- added baseDomain to spec.distribution.modules.auth in schema, required if auth.provider.type == "sso"
- regen go types from schema


nutellinoit (approved):
LGTM

ralgozino (approved):
LGTM

PR #48, opened 2023-01-20
chore: added comment to cidr_blocks to only support 1 at the moment

Al-Pragliola:
Changelist:

- improved comment on additionalFirewallrules.cidrBlocks about current support to only one rule
- improved schema, now cidrBlocks needs to have at least one element

### DIFF
--- a/templates/config/ekscluster-kfd-v1alpha2.yaml.tpl
+++ b/templates/config/ekscluster-kfd-v1alpha2.yaml.tpl
@@ -118,7 +118,7 @@ spec:
         #  - name: traffic_80_from_172_31_0_0_16
         #    # The type of the rule, can be ingress or egress
         #    type: ingress
-        #    # The CIDR blocks 
+        #    # The CIDR blocks, at the moment only one is supported, others will be ignored
         #    cidrBlocks:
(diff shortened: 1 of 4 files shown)


ralgozino:
> * improved schema, now cidrBlocks needs to have at least one element

should we set the max also until we fix the installer?

templates/config/ekscluster-kfd-v1alpha2.yaml.tpl
```diff
@@ -118,7 +118,7 @@ spec:
         #  - name: traffic_80_from_172_31_0_0_16
         #    # The type of the rule, can be ingress or egress
         #    type: ingress
-        #    # The CIDR blocks 
+        #    # The CIDR blocks, at the moment only one is supported, others will be ignored
```
ralgozino, line 121:
```suggestion
        #    # The CIDR blocks for the FW rule. At the moment the first item of the list will be used, others will be ignored. See https://github.com/sighupio/fury-eks-installer/issues/46
```

ralgozino (changes requested):

ralgozino (approved):
LGTM

PR #51, opened 2023-02-06
Feat: expand minimal conf

Al-Pragliola:


pkg/config/validation.go
```diff
@@ -19,6 +19,31 @@ const (
 var (
 	apiVersionRegex = regexp.MustCompile(apiVersionString)
 	eksVersionRegex = regexp.MustCompile(eksVersionString)
+	awsRegions      = map[string]bool{
```
ralgozino, line 22: can we get this list from some AWS API?

ralgozino (changes requested):

pkg/config/validation.go
omissis: if not, perhaps we could consider creating (and publishing?) a dedicated package for that, as it's not the first time we need this list

pkg/config/validation.go
```diff
@@ -44,6 +69,11 @@ func NewValidator() *validator.Validate {
 		return nil
 	}
 
+	err = validate.RegisterValidation("aws-region", ValidateAwsRegion)
+	if err != nil {
```
omissis, line 73: nitpick
```suggestion
	if err = validate.RegisterValidation("aws-region", ValidateAwsRegion); err != nil {
```

omissis (approved):
LGTM

pkg/config/validation.go
Al-Pragliola: added dep to package https://github.com/jsonmaur/aws-regions

ralgozino: what if we use awscli (that we already depend on) instead of adding this dependency? For example with my credentials I get these regions:

```bash
aws ec2 describe-regions | jq '.Regions[].RegionName'
"ap-south-1"
"eu-north-1"
"eu-west-3"
"eu-west-2"
"eu-west-1"
"ap-northeast-3"
"ap-northeast-2"
"ap-northeast-1"
"ca-central-1"
"sa-east-1"
"ap-southeast-1"
"ap-southeast-2"
"eu-central-1"
"us-east-1"
"us-east-2"
"us-west-1"
"us-west-2"
```

This has the advantage that the regions are the ones enabled for the account. So we also check that the region set in the configuration file is correct **and usable** for free.

ralgozino: understood, thanks. The dependency is better than hardcoded and I guess we can always improve this later if we need to.

ralgozino (approved):

PR #54, opened 2023-02-22
Fix: comment secret section in sample configuration

### LINKED ISSUE #53: Remove some properties that are not required by default
On creation of`furyctl.yml` comment the `spec.distribution.modules.ingress.nginx.tls.secret` as the default provider is `certManager`

ralgozino:
comment out the spec.distribution.modules.ingress.nginx.tls.secret section in the sample configuration file for EKSCluster because the default in the sample is using cert-manager instead of a custom certificate.

fixes #53


nutellinoit (approved):
LGTM

PR #57, opened 2023-02-27
Feature: Add custom patches

luigibarbato:
This Poc intends to add the capability to use Kustomize's custom patches within KFD's schema.  Currently the idea is to have references under the `.spec.overrides` path as below:
```
---
kind: EKSCluster
metadata:
  # The name of the cluster, will be also used as a prefix for all the other resources created on AWS
  name: example
spec:
  overrides:
    configGenerator: []
    secretsGenerator: []
    patchesJson6902: []
    patchesStrategicMerge: [] 
...
```


nutellinoit (changes requested):
customPatches (your override field) should be under the `distribution` key, not outside

we also need to check the syntax for all the fields in the template, for example, we should put a 

```
patchesStrategicMerge: | 
    {{ .value }} #with some default identation 
```

See https://github.com/sighupio/fury-distribution/blob/feature/furyctl-next/templates/distribution/manifests/auth/resources/pomerium-policy.yml.tpl#L34 as an example

patchesJson6902 should be an object array, see some example of patches json

secretGenerator and configGenerator, same as above

nutellinoit:
Proposal:

```
---
kind: EKSCluster
metadata:
  # The name of the cluster, will be also used as a prefix for all the other resources created on AWS
  name: example
spec:
  distribution:
    customPatches:
      configMapGenerator: [] <- object array
      secretsGenerator: [] <- object array
      patchesJson6902: [] <- object array
      patchesStrategicMerge: [] <- string array
```

If the files are outside the kustomize project, the `--load_restrictor LoadRestrictionsNone` argument is mandatory on `kustomize build` furyctl commands, we need to check if this argument (or feature) is still present on version 4.x and 5.x (to future proof our solution)

Following an example with "all" the permutation we use in kustomization when creating clusters:

```yaml
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources: 
  - ../../fury-kubernetes-aws/katalog/cluster-autoscaler/base


patchesStrategicMerge:
  - | 
    ---
    apiVersion: v1
    kind: Service
    metadata:
      labels:
        label1: value1
      name: cluster-autoscaler
      namespace: kube-system
  - | 
    ---
    apiVersion: v1
    kind: Service
    metadata:
      labels:
        label2: value2
      name: cluster-autoscaler
      namespace: kube-system
  - /Volumes/sighup_vol/github.com/sighupio/fury-distribution/patchtest.yaml

patchesJson6902: 
  - target:
      group: ""
      version: v1
      kind: Service
      name: cluster-autoscaler
      namespace: kube-system
    path: /Volumes/sighup_vol/github.com/sighupio/fury-distribution/patchjsontest.yaml
  - target:
      group: ""
      version: v1
      kind: Service
      name: cluster-autoscaler
      namespace: kube-system
    patch: |
          - op: add
            path: /metadata/labels/label3
            value: "value3"

configMapGenerator:
  - name: a-configmap
    files:
      - /Volumes/sighup_vol/github.com/sighupio/fury-distribution/config.example
  - name: b-configmap
    envs:
      - /Volumes/sighup_vol/github.com/sighupio/fury-distribution/env.example
      
secretGenerator:
  - name: a-secret
    files:
      - /Volumes/sighup_vol/github.com/sighupio/fury-distribution/config.example
  - name: b-secret
    envs:
      -  /Volumes/sighup_vol/github.com/sighupio/fury-distribution/env.example
```

luigibarbato:
> customPatches (your override field) should be under the `distribution` key, not outside
> 
> we also need to check the syntax for all the fields in the template, for example, we should put a
> 
> ```
> patchesStrategicMerge: | 
>     {{ .value }} #with some default identation 
> ```
> 
> See [`feature`/furyctl-next/templates/distribution/manifests/auth/resources/pomerium-policy.yml.tpl#L34](https://github.com/sighupio/fury-distribution/blob/feature/furyctl-next/templates/distribution/manifests/auth/resources/pomerium-policy.yml.tpl?rgh-link-date=2023-03-03T11%3A17%3A44Z#L34) as an example
> 
> patchesJson6902 should be an object array, see some example of patches json
> 
> secretGenerator and configGenerator, same as above

So, we should have a `customPatches` field for each module? e.g .`spec.distribution.modules.auth.pomerium.customPatches`

nutellinoit:
> So, we should have a `customPatches` field for each module? for example: `.spec.distribution.modules.auth.pomerium.customPatches`

Not right now, let's start with only the patches in the main kustomization.yaml

luigibarbato:
I used `patches` instead of `patchesJson6903` to follow https://github.com/kubernetes-sigs/kustomize/issues/2690

omissis (approved):
Tested. LGTM.

PR #74, opened 2023-03-30
docs: remove force conflicts from v1.24.0-to-v1.25.0.md

ralgozino:
remove `--force-conflicts` flag from the upgrade guide instructions. This is not needed anymore.

The `--force-conflicts` flag was needed while migrating from client-side apply to server-side.


nutellinoit (approved):
LGTM

PR #76, opened 2023-04-03
feat: update single nginx svc to use ALB controller and proxy protocol

nutellinoit:
We are using the new configuration on dual nginx, but is missing on single flavour, adding it now


omissis (approved):

ralgozino (approved):
lgtm

PR #79, opened 2023-04-04
Feat/add 1.25.1

nutellinoit:
as per title ☝️


ralgozino (changes requested):
the badge in line 122 with the link to the CNCF certification points to the 1.24 PR instead of the 1.25

PR #82, opened 2023-04-06
feat: add support for public eks clusters

omissis:
- feat(eks): improve kubernetes api access specs
- chore(eks): update code to work on public clusters


Al-Pragliola (approved):
lgtm

PR #85, opened 2023-04-13
docs: simplify compatibility matrix

### LINKED ISSUE #83: Simplify compatibility matrix with only the suggested versions for each k8s versions, and remove the older versions
Hi all, our compatibility matrix is a little bit complicated, we should simplify and remove older unsupported versions (like below 1.20/1.19)

wdy think?

PS: also, we should order them not alphabetically but with the release dates, for example 1.22.1 is more recent than 1.23.1, and have more recent packages)

ralgozino:
Closes #83

### DIFF
--- a/docs/COMPATIBILITY_MATRIX.md
+++ b/docs/COMPATIBILITY_MATRIX.md
@@ -2,8 +2,9 @@
 
 ## Maintained releases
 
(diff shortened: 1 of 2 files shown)


nutellinoit (approved):
LGTM 🚀 !

PR #88, opened 2023-05-04
feat: add checksummer tool and checksums of current tools

alessiodionisi:
SSIA


tools/checksummer/main.go
```diff
@@ -0,0 +1,124 @@
...
+			if err != nil {
+				return err
+			}
+
+			bar.Finish()
+
+			fileChecksumBytes := sha256.Sum256(fileBytes)
+			fileChecksum := hex.EncodeToString(fileChecksumBytes[:])
+
+			fmt.Printf("Checksum is %s\n", fileChecksum)
```
omissis, line 110:
```suggestion
			log.Printf("DEBUG: checksum is '%s'\n", fileChecksum)
```

omissis (changes requested):
Besides some nitpicks it looks ok, please add a bit of documentation about why is this command here and on how to use it. Also if not really necessary, I would consider dropping the bar, as it's an external dependency that probably adds little to the value of this command. jm2c

omissis (approved):
added some docs and a makefile target for user's convenience. lgtm now.

PR #94, opened 2023-05-17
feature/upgrade terraform to 1.x

omissis:
- chore: update kfd yaml
- update aws and dr modules versions in kfd yaml to point at terraform-1-compatible versions


Al-Pragliola (approved):
lgtm

PR #97, opened 2023-06-06
feature/distribution support branches release v1.25

omissis:
- feat: add checksummer tool and checksums of current tools
- address comments
- feat: add checksummer docs and makefile target
- feat: removed validation on distribution version


omissis:
We decided not to backport this to v0.25.0, closing

PR #99, opened 2023-06-06
fix/ekscluster kfd schemas release v1.25

omissis:
- fix(jsonschema): add missing aws regions
- feat: add new aws regions to the schema
- fix: amend makefile to correctly expose tools-go.
- chore: try forcing mod=readonly during go lint runs
- chore: try forcing mod=mod during go lint runs
- chore: try forcing the removal of vendor folder before lint
- chore: temporarily disable gomoddirectives linter
- chore: temporarily disable musttag linter


Al-Pragliola (approved):
LGTM

PR #107, opened 2023-06-19
feat: add yq dependency and apply script template

alessiodionisi:
SSIA

### DIFF
--- a/templates/distribution/scripts/delete.sh.tpl
+++ b/templates/distribution/scripts/delete.sh.tpl
@@ -1,5 +1,7 @@
 #!/usr/bin/env sh
 
+set -e
+
 kustomizebin="{{ .paths.kustomize }}"
 kubectlbin="{{ .paths.kubectl }}"
 yqbin="{{ .paths.yq }}"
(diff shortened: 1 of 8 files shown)


templates/distribution/scripts/delete.sh.tpl
```diff
@@ -0,0 +1,26 @@
...
+kustomizebin="{{ .paths.kustomize }}"
+kubectlbin="{{ .paths.kubectl }}"
+
+if [ "$1" = "--dry-run=true" ]; then
+  dryRun="--dry-run=client"
+else
+  dryRun=""
+fi
+
+$kubectlbin delete --all-namespaces --wait --timeout=180s --all ingress $dryRun
```
nutellinoit, line 12: Maybe the deletion should be the same between Distro and EKS provider, and we should:

* kustomize build everthing, skip CRDs and delete all the CustomResources
* Delete all statefulsets from the build
* Delete all the PVCs from our distro namespaces (logging, monitoring)
* Delete all the svc type LoadBalancers (filtered with yq)
* Delete everything else BUT CRDs
* Delete everything

(using --ignore-not-found on kubectl delete)

nutellinoit (changes requested):
One comment for a possible enhancement on the deletion

PR #112, opened 2023-07-11
fix: override modules tolerations/nodeSelectors

Al-Pragliola:


nutellinoit:
Was this tested?

Al-Pragliola:
> Was this tested?

tested with `dump template` on overrides by module or package, and also with `validate config` to check the support of `operator` in the schema.

still, another round of testing by someone else doesn't hurt

nutellinoit:
Ok, we can test them

PR #118, opened 2023-07-25
Fix: handle "-" and "." in vpn bucketNamePrefix

Al-Pragliola:
Changelist:

- added type AwsS3BucketNamePrefix to EKS schema
- now spec.infrastructure.vpn.bucketNamePrefix uses AwsS3BucketNamePrefix


g-iannelli (approved):
LGTM

PR #132, opened 2023-10-04
Release 1.25.8

g-iannelli:
What:

- fix taint regex in public schema

Depends on:
- https://github.com/sighupio/fury-eks-installer/pull/69

PR #133, opened 2023-10-04
Release 1.26.2

g-iannelli:
What:

- fix taint regex in public schema
- fix wrong version in kfd.yml for v1.26.1

Depends on:
- [ ] #132 to include the 1.25.8 release


Al-Pragliola (approved):
lgtm

PR #142, opened 2023-11-14
fix/add missing guards in kustomization yaml

omissis:
- fix: add missing guards to kustomization.yaml
- chore: bump kubectl version


Al-Pragliola (approved):
lgtm

PR #155, opened 2024-01-16
add policy module migrations when disabled

g-iannelli:
When we change the parameter on modules.policy.type from one value to another:

- kyverno to none
- gatekeeper to none
- gatekeeper to kyverno not supported
- kyverno to gatekeeper not supported


g-iannelli:
@Al-Pragliola on the rules, how can I get the starting state to configure unsopported? 
@nutellinoit if you want to move from "gatekeeper" to "kyverno", is not enougth to remove "gatekeeper" and let furyctl install "kyverno" and vice versa?

nutellinoit (approved):
LGTM

nutellinoit:
I will merge this, fixes have been done on #157

PR #162, opened 2024-01-19
feat: default policies migration

ralgozino:
Handle migrations for changing the installation of default policies in the policy module.

Add migrations for Gatekeeper and Kyverno.

Tested with an on-prem cluster:
- Ran furyctl with Gatekeeper as policy type with the default policies enabled, checked that they were working. Then disabled the default policies, ran furyctl and verified that the Constraints, Constraints Templates and Configs were deleted.
- Disabled Gatekeeper (policy type -> None) ran furyctl, re ran furyctl with Kyverno enabled with the defaults policies set to true. Checked that they were deployed and working. Then disabled the default policies and checked that they were removed from the cluster.

Fixes https://github.com/sighupio/product-management/issues/321

> [!IMPORTANT]
> Depends on https://github.com/sighupio/furyctl/pull/465 because the values to and from are booleans.

### DIFF
--- a/rules/ekscluster-kfd-v1alpha2.yaml
+++ b/rules/ekscluster-kfd-v1alpha2.yaml
@@ -3,10 +3,9 @@
 # license that can be found in the LICENSE file.
 
 ---
-
 infrastructure: []
 kubernetes: []
-distribution: 
+distribution:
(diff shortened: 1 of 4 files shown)


nutellinoit (approved):
LGTM!

PR #163, opened 2024-01-19
feat(templates/policy): install Gatekeeper monitoring if needed

ralgozino:
Deploy the monitoring options for Gatekeeper only when the monitoring module is enabled. Otherwise the installation may fail because of missing APIs from Prometheus.


nutellinoit (approved):
nice catch! LGTM

PR #160, opened 2024-01-18
Feat: add migrations for dr

nutellinoit:
This PR adds the migration for the DR module

- [x] Add migrations for OnPremises kind
- [x] Add migrations for KFDDistribution kind
- [x] Add migrations for EKSCluster kind

PS: I'm also adding all the missing e2e tests

PR #164, opened 2024-01-21
Feat: migrate logging to none

nutellinoit:
as per title

depends on #160


templates/distribution/scripts/pre-apply.sh.tpl
```diff
@@ -9,29 +9,79 @@ vendorPath="{{ .paths.vendorPath }}"
...
+
+  $kustomizebin build $vendorPath/modules/logging/katalog/logging-operated | $kubectlbin delete --ignore-not-found --wait --timeout=180s -f -
+  $kustomizebin build $vendorPath/modules/logging/katalog/configs | $kubectlbin delete --ignore-not-found --wait --timeout=180s -f -
+
+  $kustomizebin build $vendorPath/modules/logging/katalog/minio-ha > delete-logging-minio-ha.yaml
+
+{{- if eq .spec.distribution.modules.monitoring.type "none" }}
+  if ! $kubectlbin get apiservice v1.monitoring.coreos.com; then
+    cat delete-logging-minio-ha.yaml | $yqbin 'select(.apiVersion != "monitoring.coreos.com/v1")' > delete-logging-minio-ha-filtered.yaml
+    cp delete-logging-minio-ha-filtered.yaml delete-tracing-minio-ha.yaml
```
ralgozino, line 75: I see you use this approach in several places, is there any reason for not using the `--inplace` flag of `yq`? I think it would be better (1 command instead of 2 and with the `cp` approach you may have a typo in the destination filename that with in place you are sure is the same.

templates/distribution/scripts/pre-apply.sh.tpl
```diff
@@ -9,29 +9,79 @@ vendorPath="{{ .paths.vendorPath }}"
...
   $kubectlbin delete --ignore-not-found --wait --timeout=180s ingress -n pomerium cerebro
-  $kustomizebin build $vendorPath/modules/logging/katalog/opensearch-dashboards | $kubectlbin delete --ignore-not-found --wait --timeout=180s -f -
-  $kustomizebin build $vendorPath/modules/logging/katalog/opensearch-triple | $kubectlbin delete --ignore-not-found --wait --timeout=180s -f -
-  $kustomizebin build $vendorPath/modules/logging/katalog/cerebro | $kubectlbin delete --ignore-not-found --wait --timeout=180s -f -
-  echo "Waiting 3 minutes"
-  sleep 180
-  $kubectlbin delete -l app.kubernetes.io/name=opensearch pvc -n logging --wait --timeout=180s
+
+  $kustomizebin build $vendorPath/modules/logging/katalog/opensearch-dashboards > delete-opensearch-cerebro.yaml
+  $kustomizebin build $vendorPath/modules/logging/katalog/opensearch-triple >> delete-opensearch-cerebro.yaml
```
ralgozino, line 34: what if the user has selected the single version? will this work anyway or should we consider that?

ralgozino (commented):
overall looks good to me, I left some questions before approving

templates/distribution/scripts/pre-apply.sh.tpl
nutellinoit: I was not able to make it work with yq, maybe I did something wrong

templates/distribution/scripts/pre-apply.sh.tpl
nutellinoit: Since we are deleting, triple has only some environment variables and the replica set at 3, for a deletion it does not matter

PR #180, opened 2024-02-13
feat(ci/e2e): run Kind on baremetal

ralgozino:
Stop using dedicated VMs on vmWare for running the ditro's end-to-end tests and use a Kind cluster directly on drone's worker node instead.

This greatly improves the performance. Using the previous approach, e2e tests took ~1h, with this approach it takes ~30m.

Downsides:
- even though the control-plane for the Kind cluster will listen on the node's loopback interface (so it will NOT be exposed to the Internet) because it uses the host docker daemon, the e2e tests step must run in the worker's network to reach the control-plane.

resolves https://github.com/sighupio/product-management/issues/345


nutellinoit (approved):
LGTM!!! 🚀

PR #185, opened 2024-02-28
feat: moved eks infra/kube terraform folders from furyctl

Al-Pragliola:


nutellinoit (approved):
LGTM, same as the furyctl PR, we need to test all the changes together

alessiodionisi (approved):

PR #186, opened 2024-03-01
feat: added security parameters in the OnPremises schema

stefanoghinelli:
This PR includes the encryption parameters in the OnPremises schema.

Related to: [product-management/issues/346](https://github.com/sighupio/product-management/issues/346)


alessiodionisi (approved):

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -90,6 +90,9 @@
         },
         "advanced": {
           "$ref": "#/$defs/Spec.Kubernetes.Advanced"
+        },
+        "encryption": {
```
nutellinoit, line 94: The correct place for encryption should be inside the advanced key, not at the same level

nutellinoit (changes requested):
Just a change, encryption should be under "advanced" key

schemas/public/onpremises-kfd-v1alpha2.json
stefanoghinelli: Moved field and tested again

PR #187, opened 2024-03-01
fix: recucer not working switching from none to mimir with externalEndpoint

Al-Pragliola:
partially fixes: https://github.com/sighupio/furyctl/issues/487


FedericoAntoniazzi (commented):
I tested with the reported case and can confirm it worked

FedericoAntoniazzi (approved):

alessiodionisi (approved):

PR #189, opened 2024-03-15
Feat: add additional static clients on dex

nutellinoit:
This PR adds a new parameter on dex configuration `additionalStaticClients` and can be used like this:

```
        dex: 
          connectors:
            - type: ldap
              ...
          additionalStaticClients:
            - id: test
              redirectURIs:
                - https://argocd.test/auth/callback
                - https://argocd.test/auth/login
              name: 'ArgoCD Login'
              secret: XXXXXX
```

I also update the kfd.yaml with the new on premises installer for compatibility purposes

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -159,7 +159,7 @@ steps:
     environment:
       CLUSTER_NAME: ${DRONE_REPO_NAME}-${DRONE_BUILD_NUMBER}
       KUBECONFIG: /drone/src/kubeconfig
-      FURYCTL_VERSION: v0.27.3-rc.2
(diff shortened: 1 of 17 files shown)


ralgozino (dismissed):
tested locally, does what expected <3

the only thing to keep in mind is that this doesn't do any validation that the _content_ of `additionalStaticClients` is correct.

nutellinoit:
> the only thing to keep in mind is that this doesn't do any validation that the _content_ of `additionalStaticClients` is correct.

Exactly, the logic is the same as connectors

Al-Pragliola (approved):
lgtm

PR #193, opened 2024-03-27
chore(templates/on-prem): fix usage of deprecated flag in drain command

ralgozino:
stop using the `--delete-local-data` flag for the drain command that has been deprecated in 2020 and use `--delete-emptydir-data` instead.

Ref: https://github.com/kubernetes/kubernetes/pull/95076


nutellinoit (approved):
LGTM!

PR #199, opened 2024-04-04
feat(on-premises): set nodes roles with the node pool name

ralgozino:
Use the node pool name defined in the furyctl.yaml as the well-knonw label used to define the "role" by kubectl.

Only for nodes that are not the control-plane because the control-plane role is set by kubeadm.

This will be executed as part of the Kubernetes phase and after the creation of the cluster.


nutellinoit (approved):
LGTM!

PR #202, opened 2024-04-04
feat(templates/monitoring): use kubeadm-sm for on-prem

ralgozino:
use kubeadm-sm instead of eks-sm when kind is OnPremises.

Note: we don't access (yet) to the `.kind` key of the configuration file from the templates, so we use `common.provider` instead. This is set to "eks" when the kind is EKSCluster.


nutellinoit (approved):
LGTM!

PR #203, opened 2024-04-04
fix(templates/ingress): always apply nodeSelector and Tolerations for certmanager

ralgozino:
certmanager is always installed as a dependency of the distribution, so we always apply nodeSelectors and Tolerations to certmanager instead of checking if ingress.tls.type is certmanager.


nutellinoit (approved):
LGTM!

PR #204, opened 2024-04-04
fix(templates/monitoring): add nodeSelectors and Tolerations to mimir

ralgozino:
Add missing nodeSelector and Tolerations to mimir, so it is deployed in the desired nodes like the rest of the infra components.


nutellinoit (approved):
LGTM!

PR #205, opened 2024-04-10
fix(template/distribution/monitoring/mimir): set max_global_series_per_user to unlimited

ralgozino:
Mimir limits by default the amount of time series a user can push to 150000 to protect from a single user DoSing the service.

In our installation, we have just one "fury" user, used by Prometheus to push all the time series to Mimir. So, as a maximum we will push all the time series available in Prometheus, making it a practical limit.

In a "small" KFD installation (without all the modules) we are already very close to or over the default 150000 limit.


nutellinoit (approved):
LGTM!

Al-Pragliola (approved):

PR #212, opened 2024-04-23
Feat: release 1.29.0

nutellinoit:
As per title, this PR adds the support for the new 1.29 version


kfd.yaml
Al-Pragliola, line 1: version: v1.28.0 -> version: v1.29.0

Al-Pragliola (changes requested):

ralgozino (changes requested):
what Alessio said. The rest is looking good

kfd.yaml
nutellinoit: fix'd

ralgozino (approved):

PR #214, opened 2024-04-30
fix(templates/auth): make websockets work with Pomerium

ralgozino:
Add needed configuration to Pomerium policy so websockets in Grafana and Minio ingresses work.


ralgozino:
Marking as draft until we confirm that this is not an issue with my test environment and is a general problem

ralgozino:
we confirmed with @smerlos that the issue is present and that this fix works as expected

nutellinoit (approved):
LGTM!

PR #215, opened 2024-04-30
fix(templates/pomerium): disable Pomerium debug logs

ralgozino:
set log level to info in pomerium instead of debug. Debug logs should not be enabled by default.


nutellinoit (approved):
LGTM!

PR #223, opened 2024-06-07
Add Forecastle groups in ingress with SSO enabled

sbruzzese902:
Hello guys 👋 
This proposal aims to add the annotation `forecastle.stakater.com/group` to ingresses when we use Pomerium.

In this way we can have a web view similar to this one, even if all ingresses are in the same namespace.
Screenshot 2024-06-07 at 11 38 19

I hope you like it 😄


ralgozino:
Hi Simo! I like the idea, but what is the difference with the current grouping by Namespace? For you specific example it does not change much does it?

nutellinoit:
Using Pomerium, all the ingresses are under "pomerium" namespace, this change make them under a logic group instead of an "hard" one, I think

sbruzzese902:
Exactly, @nutellinoit 😄

ralgozino:
riiight, I forgot about that <3

nutellinoit (approved):
To be tested after merge

PR #230, opened 2024-07-04
feature(networking): add internal Hubble ingress

ralgozino:
- add ingress manifests for Hubble UI behind pomerium when auth is enabled
- update networking schema to allow overriding defaults
- update schemas documentation
- update go-models
Note: this does not update EKS schema because cilium is not an option there.

fixes https://github.com/sighupio/fury-kubernetes-networking/issues/81


nutellinoit (approved):
To be tested after merge

PR #231, opened 2024-07-05
feat: make schema docs generated

Al-Pragliola:


Al-Pragliola:
> LGTM ❤️
> 
> How do you imagine the workflow with this? I'm thinking somethig like:
> 
> * Each PR updates the schema(s) file(s) only (if the PR is changing something in the schema)
> * The PR gets merged into `main`
> * Before doing the release of a new version of the distro, somebody runs the `md-gen` tool and updates the docs, maybe with another PR.

yep, it's exactly as I imagine the new workflow, if we want we can also enforce it with linting in CI when releasing (doing a diff of the md-gen and the actual schema docs)

alessiodionisi (approved):

PR #232, opened 2024-07-08
feature(onpremises): support containerd mirrors

ralgozino:
this patch makes the containerd configuration template more flexible to support configuring registry mirrors without needing to set also the authentication, auth details where previously always required otherwise the templating would fail.

> [!NOTE]
> This PR needs a fix to the on-prem installer included in this another PR:
> - https://github.com/sighupio/fury-kubernetes-on-premises/pull/87


nutellinoit (approved):
to be tested after merge

PR #236, opened 2024-07-17
feature(onprem): enable haproxy monitoring

ralgozino:
enable metrics scraping from the haproxy loadbalancers and deploy the dashboards and alerts for haproxy from the monitoring module.

Note: I was unsure on setting some auth on the prometheus endpoint. Being that the 8405 port most likely will be accessible only from the inside I think that is enough. But let me know what you think.

> [!WARNING]
> This PR depends on the PR that adds the `haproxy` package to the monitoring module:
> - https://github.com/sighupio/fury-kubernetes-monitoring/pull/179
>
> And it is also has been rebased on the following PR that introduces PrometheusAgent mode that we needed to consider:
> - #228


ralgozino:
setting this PR as draft. I'll rebase these changes on top of the PR #228 because we'll need to consider the cases introduced by that PR

ralgozino:
I've rebased this PR to be branched from the changes introduced by #228 so we could consider the case when monitoring is in PrometheusAgent mode.

It is now ready for review.

PR #238, opened 2024-07-25
feature(templates): YAML language server schema autompletion

ralgozino:
add directive to the configuration templates to point the YAML LSP language server to the right schema, so you'll get validation and autocomplete in your editor. At least until we publish the schemas on schema.org.


nutellinoit (approved):
LGTM!

PR #242, opened 2024-08-01
fix(templates/monitoring): don't apply grafana config when type is pr…

ralgozino:
…ometheusAgent

Grafana config secret must only be created when grafana is being installed (monitoring types prometheus and mimir, not when it is none or prometheusAgent)


nutellinoit (approved):
LGTM!

PR #271, opened 2024-10-04
fix: custom registry onpremises configuration

alessiodionisi:
- Added `.spec.kubernetes.advanced` field check

### DIFF
--- a/templates/kubernetes/onpremises/hosts.yaml.tpl
+++ b/templates/kubernetes/onpremises/hosts.yaml.tpl
@@ -75,11 +75,12 @@ all:
         {{- if index .spec.kubernetes.advanced.oidc "group_prefix" }}
         oidc_group_prefix: "{{ .spec.kubernetes.advanced.oidc.group_prefix }}"
         {{- end }}
-
         {{- end }}
 
+        {{- if index .spec.kubernetes "advanced" }}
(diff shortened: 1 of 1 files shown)


stefanoghinelli (approved):
lgtm

ralgozino (approved):
an improvement that we could do is to set the default value of `.spec.kubernetes.advanced` to `{}` and we can get rid of a lot of if clauses in this file

alessiodionisi:
@ralgozino if I remember correctly we can set defaults only for distribution values with the current template system implementation

ralgozino:
> @ralgozino if I remember correctly we can set defaults only for distribution values with the current template system implementation

I'm not sure I understand. What I ment was that , if we set `.spec.kubernetes.advanced` to an empty object `{}` in the `defaults/onpremises-kfd-v1alpha2.yaml` file like we do for `customPatches` for example, we don't need to check if it is defined in the template.

alessiodionisi:
> > @ralgozino if I remember correctly we can set defaults only for distribution values with the current template system implementation
> 
> I'm not sure I understand. What I ment was that , if we set `.spec.kubernetes.advanced` to an empty object `{}` in the `defaults/onpremises-kfd-v1alpha2.yaml` file like we do for `customPatches` for example, we don't need to check if it is defined in the template.

Unfortunately defaults files are applied only to `spec.distribution` furyctl configurations.

PR #274, opened 2024-10-14
[BREAKING] fix(policy/kyverno)!: use `Enforce` and `Audit` instead of all lowercase

ralgozino:
Changes the possible values for the `validationFailureAction` field of Kyverno from `audit` and `enforce` to `Audit` and `Enforce`.

Previously we were using `audit` and `enforce`, all in lowercase, as the possible values. These values were triggering the following deprecation warning:

```
Warning: Validation failure actions enforce/audit are deprecated, use
Enforce/Audit instead.
```

After this change, the possible values will match the ones expected by Kyverno.

BREAKING CHANGE: users that have Kyverno enabled and have manually set the `validationFailureAction` value will need to adjust it, otherwise the schema validation fails with:

```
ERRO error while validating configuration file: error while validating
against schema: jsonschema:
'/spec/distribution/modules/policy/kyverno/validationFailureAction' does
not validate with
file:///var/folders/lb/zscvwt_s6fx6qdrfv4g51dsc0000gn/T/furyctl-1755893597/data/schemas/public/onpremises-kfd-v1alpha2.json#/properties/spec/$ref/properties/distribution/$ref/properties/modules/$ref/properties/policy/$ref/properties/kyverno/$ref/properties/validationFailureAction/enum:
value must be one of "Audit", "Enforce"
```

### DIFF
--- a/defaults/ekscluster-kfd-v1alpha2.yaml
+++ b/defaults/ekscluster-kfd-v1alpha2.yaml
@@ -229,7 +229,7 @@ data:
       kyverno:
         # this configuration adds namespaces to the excluded list, actually whitelisting them
         additionalExcludedNamespaces: []
-        validationFailureAction: enforce
+        validationFailureAction: Enforce
         installDefaultPolicies: true
     # dr module configuration
     dr:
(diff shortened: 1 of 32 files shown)


nutellinoit (approved):
LGTM, let's put a [breaking] description in the PR title, just to remember when we create release notes that this is breaking

ralgozino:
> LGTM, let's put a [breaking] description in the PR title, just to remember when we create release notes that this is breaking

according to semantic commits that's what the `!` after the `feat(*)` indicates, but I agree that is not very obvious. I edited the title.

PR #275, opened 2024-10-14
fix(on-prem): use the `.spec.kubernetes.advanced.users.org` parameter for kubernetes users creation

### LINKED ISSUE #265: on-prem: `kubernetes.advanced.users.org` seems to be ignored
Even though the schema allows setting the `kubernetes.advanced.users.org` parameter to create new users via kubeadm and associating them to an organization, all the kubeconfig files created using the `kubernetes.advanced.users.names` list belong to the `sighup` organization after inspecting the created certificates, instead of the one specified in the `org` field.

I did not find traces of the value of this parameter being used.

speziato:
Closes #265

### DIFF
--- a/templates/kubernetes/onpremises/hosts.yaml.tpl
+++ b/templates/kubernetes/onpremises/hosts.yaml.tpl
@@ -50,6 +50,9 @@ all:
         kubernetes_users_names:
 {{ .spec.kubernetes.advanced.users.names | toYaml | indent 10 }}
         {{- end }}
+        {{- if index .spec.kubernetes.advanced.users "org" }}
+        kubernetes_users_org: "{{ .spec.kubernetes.advanced.users.org }}"
+        {{- end }}
         {{- end }}
 
         {{- if and (index .spec.kubernetes "advanced") (index .spec.kubernetes.advanced "oidc") }}


nutellinoit (commented):
Did you test this change?

ralgozino (changes requested):
Unfortunately this change is not working for me.

This is what I did:

1. Cleaned up local files and the `.furyctl` folder.
2. Added a user to the list in the furyctl.yaml file under `.spec.kubernetes.advanced.users.names` and set `.spec.kubernetes.advanced.users.org` to a custom value.
3. Run `furyctl apply --phase kubernetes`
4. I checked the `<username>.kubeconfig` file that gets copied in the current folder for the Organization property in the certificates with the following command:

```bash
yq '.users[0].user.client-certificate-data' <username>.kubeconfig | base64 -d | openssl x509 -text | grep Subject:
```

I got `O=sighup, CN=<username>` as value.

I expected the additional user to have the `O=<org I set in the furyctl.yaml>`.

ralgozino (approved):
As you wisely pointed out to me, this change was being overridden by furyctl compatibility patches in my test 🤕  .

Re testing these changes without the compatibility patches, the organization is set properly for the additional users.

well done @speziato 🎉

speziato:
@nutellinoit 
> Did you test this change?

Yep, but I tested it in a wrong way: I edited the compatibility patch in furyctl and built it locally to test with KFD v1.29.4, saw that it was working and pushed the change here. The test I did was not as good as @ralgozino's one. 

Basically, to test locally you can just clone this repo, checkout this branch, change the  `version` property in `kfd.yaml` to something for which furyctl does not have embedded compatibility patches (such as `v1.29.5`), create a `furyctl.yaml` with that inexisting version and run `furyctl create --distro-path <path-of-the-cloned-repo>`

```bash
git clone https://github.com/sighupio/fury-distribution --branch=fix/on-prem-user-org
cd fury-distribution
sed -I "" -r 's#^(version:\ ).*#\1v1.29.5#g' kfd.yaml
furyctl create config --kind OnPremises --version v1.29.5 --distro-location .
# add the advanced.users.names and advanced.users.org properties to the furyctl.yaml
furyctl apply --distro-location . --phase kubernetes
yq '.users[0].user.client-certificate-data' <username>.kubeconfig | base64 -d | openssl x509 -text | grep Subject:
```

nutellinoit (approved):
LGTM!

PR #277, opened 2024-10-21
fix(auth): make Hubble Ingress work with basicAuth

### LINKED ISSUE #276: Hubble Ingress not working when Auth Provider is Basic Auth
The Hubble ingress returns an error 503 when the `auth.Provider` is `basicAuth`.

The error is due to the secret `basic-auth` not existing in the `kube-system` namespace where the ingress is created.

ralgozino:
Add missing secret for making basicAuth work for the Hubble Ingress created in the kube-system namespace when Networking type is cilium.

Fixes #276


ralgozino:
Maybe we should apply this fix as a compatibility patch with furyctl too for previous versions.

alessiodionisi (approved):

PR #278, opened 2024-10-22
feat(on-premises): allow clusters without nodes

ralgozino:
Allow creating on-premises clusters with only control-plane hosts and no additional nodes, for minimal cluster installations.


nutellinoit (approved):
LGTM! 🚀

PR #279, opened 2024-10-22
fix: do not install x509 data planes on EKS

nutellinoit:
This component is not needed on EKS and it will just give false alerting


ralgozino (approved):
changes lgtm

I suppose you checked but just to double check, should we also skip the control-plane daemonset in EKS?

nutellinoit:
> changes lgtm
> 
> I suppose you checked but just to double check, should we also skip the control-plane daemonset in EKS?

I didn't skip it because it will never be deployed on worker nodes by design

PR #280, opened 2024-10-22
fix(templates/scripts): auth migrations

ralgozino:
Fix auth migrations:
- Consider Hubble Ingress
- basicAuth -> sso was defined twice, so it was not working properly and sso->basicAuth was not working at all.


nutellinoit (approved):
LGTM! Nice catch!

PR #281, opened 2024-10-22
fix(templates/ingress): nginx type none and tls secret, migration to none

ralgozino:
- Don't create `ingress-nginx-global-tls-cert` secret in `ingress-nginx` namespace when `nginx.tls.provider` is set to `secret` and `nginx.type: none` at the same time, it results in an error bacause the namespace does not exist.
- Migration for `nginx.type -> none`: ingresses in `kube-system` namespace (dex, gangway, hubble) were not being deleted. Delete all the ingresses in the `kube-system` namespace as part of the migration.

> [!NOTE]
> This PR is branched off #280 because it modifies some of the same files. You may see changes from both PRs here until 280 is merged.

### DIFF
--- a/templates/distribution/manifests/ingress/kustomization.yaml.tpl
+++ b/templates/distribution/manifests/ingress/kustomization.yaml.tpl
@@ -37,7 +37,7 @@ resources:
   - resources/ingress-infra.yml
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


nutellinoit (approved):
LGTM!

alessiodionisi (dismissed):

PR #283, opened 2024-10-28
fix: correct dns validation when nginx is single, dual or none

### LINKED ISSUE #86: Furyctl(next) - Creating a cluster with single nginx dns private is needed even if it's not used
Example error:

```
-> ./furyctl validate config                                                                            
INFO Downloading distribution...                  
ERRO error while validating against schema: jsonschema: '/spec/distribution/modules/ingress/dns' does not validate with https://schema.sighup.io/kfd/ekscluster-kfd-v1alpha2-public.json#/properties/spec/$ref/properties/distribution/$ref/properties/modules/$ref/properties/ingress/$ref/properties/dns/$ref/required: missing properties: 'private' 
ERRO configuration file validation failed  
```

Using distribution `v1.25.2-rc.0` and furyctl `0.25.0-alpha.1`

alessiodionisi:
Closes #86.

### DIFF
--- a/schemas/public/ekscluster-kfd-v1alpha2.json
+++ b/schemas/public/ekscluster-kfd-v1alpha2.json
@@ -1108,29 +1108,83 @@
       },
       "required": [
         "baseDomain",
-        "dns",
         "nginx"
       ],
-      "if": {
-        "properties": {
-          "nginx": {
+      "allOf": [
+        {
+          "if": {
             "properties": {
-              "tls": {
+              "nginx": {
                 "properties": {
-                  "provider": {
-                    "const": "certManager"
+                  "type": {
+                    "const": "dual"
                   }
                 }
               }
             }
+          },
+          "then": {
+            "required": [
+              "dns"
+            ],
+            "properties": {
+              "dns": {
+                "required": [
+                  "public",
+                  "private"
+                ]
+              }
+            }
+          }
+        },
+        {
+          "if": {
+            "properties": {
+              "nginx": {
+                "properties": {
+                  "type": {
+                    "const": "single"
+                  }
+                }
+              }
+            }
+          },
+          "then": {
+            "required": [
+              "dns"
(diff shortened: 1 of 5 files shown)


ralgozino (commented):
Hi Ale, I see you asked for review but this PR is still draft. Is it ready?

Also, have you tested these changes?

alessiodionisi:
Yes, it's a work in progress. Sorry for the review notification!

alessiodionisi:
I did some tests on AWS and it seems to work in all cases. Can you do some checks too?

ralgozino (commented):
code wise looks good to me. I'll left the approve to @nutellinoit that reported the original issue.

ralgozino:
maybe @g-iannelli can take a look :)

g-iannelli:
Is the purpose of the PR to avoid adding the `dns` key when nginx is `none` and in the case of `single` to only require the `public`?

alessiodionisi:
Yes!

schemas/public/ekscluster-kfd-v1alpha2.json
```diff
@@ -1108,29 +1108,83 @@
...
+            }
+          },
+          "then": {
+            "required": [
+              "dns"
+            ],
+            "properties": {
+              "dns": {
+                "required": [
+                  "public"
```
g-iannelli, line 1159: I don't know if it makes sense to allow users to choose between public and private in the case of a single ingress class. Maybe currently, we don't have any cases in which there is just an ingress class, and it should be private.

g-iannelli (approved):
LGTM

schemas/public/ekscluster-kfd-v1alpha2.json
alessiodionisi: Yes, currently to have the private one, the dual ingress must be enabled.

PR #284, opened 2024-10-28
fix: adding a new field on plugins spec to permit a new field

nutellinoit:
This is needed to disable helmfile validation (helm diff command) during installation, when installin CR that depends on CRD deployed by the chart itself (for example, the netapp trident driver)

### DIFF
--- a/schemas/public/spec-plugins.json
+++ b/schemas/public/spec-plugins.json
@@ -87,6 +87,10 @@
               "type": "string"
             },
             "description": "The values of the release"
+          },
+          "disableValidationOnInstall": {
+            "type": "boolean",
+            "description": "Disable helm diff validation on install"
           }
         },
         "required": ["name", "namespace", "chart"]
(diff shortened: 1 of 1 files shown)


speziato (dismissed):
LGTM!

schemas/public/spec-plugins.json
```diff
@@ -87,6 +87,10 @@
               "type": "string"
             },
             "description": "The values of the release"
+          },
+          "disableValidationOnInstall": {
+            "type": "boolean",
+            "description": "Disable helm diff validation on install"
```
ralgozino, line 93:
```suggestion
            "description": "Disable running `helm diff` validation when installing the plugin, it will still be done when upgrading."
```

ralgozino (changes requested):
a minor tweak to the description, I THINK it is correct (it is not 100% clear to me from Helmfile docs)

ralgozino (approved):
lgtm! we tested this live 👯

PR #288, opened 2024-11-08
fix eks terraform template with missing ami_owners

g-iannelli:
As the `ami_owners` has been introduced in https://github.com/sighupio/fury-eks-installer/pull/80 now it's possible to use the value of `nodePools.ami.owner` in the eks installer


nutellinoit (approved):
LGTM!

PR #289, opened 2024-11-08
fix(templates/monitoring-auth): use pomerium as JWT endpoint

ralgozino:
According to https://www.pomerium.com/docs/guides/grafana#configure-grafana the URL for the `jwks.json` file should be Pomerium's URL and not Grafana's.

I think now works because Grafana's URL also passes through Pomerium anyway, but better to have this properly configured to avoid possible issues in the future.


stefanoghinelli (approved):
Great observation. LGTM!

alessiodionisi (approved):

alessiodionisi:
Good catch!

PR #291, opened 2024-11-13
fix(templates/auth/pomerium): route for minio-tracing missing

ralgozino:
Fix login in Pomerium routes definition, minio-tracing route was not being added when logging is set to `none`.


stefanoghinelli (approved):
lgtm

PR #292, opened 2024-11-13
feat/loki-distributed-patch

lukeskywlkr:
This PR introduces a way to migrate existing clusters towards loki tsdb store and schema v13. Since version 2.8 Loki introduced a new single store option that has become the default store for all new Loki deployments, the TSDB store. Along with TSDB a new schema version has been introduced (v13) which enables support to structured metadata. So it introduces a breaking schema change. Below there is a detailed explanation and some examples:

## Details ⚙️ 
- **New option for Logging:**
The Loki configuration has been extended to accommodate the new `tsdbSchemav13Migration` option to allow a migration towards TSDB and schema v13 (note: **this is a breaking change**):
  ```yaml
  ...
    loki:
      backend: minio
      externalEndpoint:
        endpoint: s3.eu-west-1.amazonaws.com
        insecure: false
        secretAccessKey: example
        accessKeyId: example
        bucketName: lokibucket
      tsdbSchemav13Migration:
        enabled: true
        schemaConfig:
          tsdbStartDate: "2024-11-18"
    customOutputs: {}
  ...
  ```
  The `tsdbSchemav13Migration` has two option:
  - `enabled` (**required**): tells to furyctl that the migration have to be executed
  - `schemaConfig`: configures details for the schema config for the purpose of the migration

  As we can see in the example above `schemaConfig` has only an option: `tsdbStartDate`. This should be a string in `ISO 8601` date format and it represents the day starting from which Loki will record logs with the new store and schema.

  ℹ️ **Note**:  Loki will assume the start of the day on the UTC midnight of the specified day.
## Breaking Changes 💔 
- **Loki store and schema change:**
  A new store and schema has been introduced in order to improve efficiency, speed and scalability of Loki clusters.


lukeskywlkr:
Closed and moved to #306

PR #293, opened 2024-11-13
fix(migrations): remove minio-monitoring ingress when migrating `mimir` -> `none`

ralgozino:
Remove `minio-monitoring` ingress when switching from `.spec.distribution.modules.monitoring.type: mimir` to `.spec.distribution.modules.monitoring.type: none`.


stefanoghinelli (approved):
lgtm

PR #294, opened 2024-11-13
fix(minio): more specific tls secret name

sbruzzese902:
This PR is to better specify minio's tls secret name.

Now the secret is called `minio-tls` for all the variants: logging, monitoring and tracing. Most of the time this is not a problem.

The issue that we get as of today is when auth is `sso`, so all ingresses are deployed in `pomerium` namespace and in this case they are sharing all the same tls secret (and certificate).

Changing all the names solves the issue.


ralgozino (approved):
lgtm, nice catch!

PR #296, opened 2024-11-14
fix(templates): fix ingress tlsSecret name generation

ralgozino:
- Fix a bug introduced in #294 that broke templates for ingresses.
- Change the logic to generate the TLS secretName, instead of using the package name use the prefix (the first part of the ingress hostname).

This will change the TLS secretName for the several MinIOs ingresses that we have from `minio` to `minio-ingress`, `minio-logging`, `minio-monitoring`.

As a side effect, Forecastle's TLS secretName will also be changed, from `forecastle-tls` to `directory-tls`.

I think it is actually better this way than what we had originally, i.e. is better to use the prefix to generate the tlsSecretName than the package name, but if there are downsides that I'm not seeing we could revert to how it was originally (but the several minio ingresses will be broken)

### DIFF
--- a/templates/distribution/_helpers.tpl
+++ b/templates/distribution/_helpers.tpl
@@ -125,7 +125,7 @@
     - hosts:
       - {{ template "ingressHost" . }}
     {{- if eq .spec.distribution.modules.ingress.nginx.tls.provider "certManager" }}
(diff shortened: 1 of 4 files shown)


nutellinoit (approved):
LGTM

PR #301, opened 2024-11-15
Fix/migrations yaml

sbruzzese902:
I've fixed the generation of `yaml`s in 2 points:

- external DNS deletion
- opensearch deletion

When two or more kustomize builds are combined in the same yaml file, if the first one does not end with `---`, some resources are not deleted correctly.

Please find attached examples before the fix:
- [delete-opensearch.yml.txt](https://github.com/user-attachments/files/17778220/delete-opensearch.yml.txt) lines 196-197
- [delete-external-dns.yaml.txt](https://github.com/user-attachments/files/17778233/delete-external-dns.yaml.txt) lines 135-136

I think in general we should avoid the combination of kustomize builds or always add a `---` between them.


nutellinoit:
nice catch!

nutellinoit (approved):
LGTM!

PR #302, opened 2024-11-15
Feat/add network policies

sbruzzese902:


nutellinoit (commented):
Overall this PR is golden ❤️ 

I have some questions, on some components/rules (note: I did not try to understand all of the network policies):

1. Can the ingress controller do egress everywhere? (I would like to have no blockers here to use ingress on other namespaces (or outside of the cluster))
2. Can cert-manager do egress everywhere? (In case of DNS challenge, for example)
3. Can external-dns do egress everywhere? (same as above)
4. Can loki do egress everywhere? (for example we are using an external S3 storage)
5. Can fluentd do egress everywhere? (if we install a different flow output in the cluster, for example)
6. Can alert manager do egress everywhere? (It needs to connect to arbitrary endpoints to send alerts)
7. Can prometheus do egress everywhere? (to scrape, but also to send metrics to a different location)
8. Can mimir do egress everywhere? (to save data on S3 for example)
9. Can tempo connect egress everywhere and also accept data (ingress) from everywhere?

Thank you

sbruzzese902:
1. Yes
2. `cert-manager` can do egress towards port 53 TCP/UDP, 80 TCP and 443 TCP globally (we followed [this](https://cert-manager.io/docs/installation/best-practice/#network-requirements) )
3. It seems we forgot `external-dns` 😢, let us fix it!
4. This needs to be added. We only created `loki`=>`minio` rule.
5. At the moment `fluentd` can only perform egress towards `minio` and `opensearch` OR `loki`.
6. Yes
7. Yes
8. `mimir` can perform egress globally, if its backend is not `minio`.
9. `tempo` backend is the same as `mimir`, but it accepts traffic only from `grafana`.
10. `pomerium` can only communicate with the known services from the distro (e.g. `forecastle`, `grafana`...).

We can be more permissive on the traffic, so we avoid breaking stuff or forcing the creation of many network policies.

nutellinoit:
> We can be more permissive on the traffic, so we avoid breaking stuff or forcing the creation of many network policies.

Yes please, we need balance between security and ease of use

sbruzzese902:
Everything was changed as we agreed, including the diagrams 😄

PR #306, opened 2024-11-18
feat(logging): update logging v4.0.0 rc

lukeskywlkr:
This PR introduces a way to migrate existing clusters towards loki tsdb store and schema v13. Since version 2.8 Loki introduced a new single store option that has become the default store for all new Loki deployments, the TSDB store. Along with TSDB a new schema version has been introduced (v13) which enables support to structured metadata. So it introduces a breaking schema change. Below there is a detailed explanation and some examples:

## Checklist for Testing 🧪 

- [x] Tested schema on furyctl.yaml file with date format
- [x] Tested new Loki installation
- [x] Tested reading old logs with old format from S3 compatible storage
- [x] Verified that changing the TSDB schema on the current date on an existing storage will make the current day work only with the new tsdb storage
- [x] Tested migration to opensearch
- [x] Tested opensearch index patterns
- [x] Tested migration from opensearch to loki and verified the fix to the template if tsdbStartDate is null

## Details ⚙️ 
- **New option for Logging:**
The Loki configuration has been extended to accommodate the new `tsdbStartDate` option to allow a migration towards TSDB and schema v13 (note: **this is a breaking change**):
  ```yaml
  ...
    loki:
      tsdbStartDate: "2024-11-18"
  ...
  ```
  - `tsdbStartDate`  (**required**): configures details for the schema config for the purpose of the migration

  `tsdbStartDate` should be a string in `ISO 8601` date format and it represents the day starting from which Loki will record logs with the new store and schema.

  ℹ️ **Note**:  Loki will assume the start of the day on the UTC midnight of the specified day.

## Breaking Changes 💔 

- **Loki store and schema change:**
  A new store and schema has been introduced in order to improve efficiency, speed and scalability of Loki clusters.


schemas/private/ekscluster-kfd-v1alpha2.json
```diff
@@ -1192,10 +1206,18 @@
             }
           }
         },
+        "tsdbStartDate": {
+          "type": "string",
+          "format": "date",
+          "description": "The date loki have to switch to TSDB and schema v13"
```
ralgozino, line 1212: Please remember that this field will end up:
- in the autocomplete of your code editor that pulls the schema to validate the file structure and contents.
- In KFDs documentation site, for example here: https://docs.kubernetesfury.com/docs/furyctl/providers/onpremises#specdistributionmodulesloggingloki

So, let's try to be a little more expressive and provide useful information to the user that has to fill this field. Right now is kind of the same as the key but with more words. I suggest something like:

```suggestion
          "description": "Starting from versions 1.28.4, 1.29.5 and 1.30.0 of KFD, Loki will change the time series database from BoltDB to TSDB and the schema from v11 to v13 that it uses to store the logs.\n\nThe value of this field will determine the date when Loki will start writing using the new TSDB and the schema v13, always at midnight UTC. The old TSDB and schema will be kept until they expire for reading purposes.\n\nValue must be a string in `ISO 8601` date format (`yyyy-mm-dd`). Example: `2024-11-18`."
```

Feel free to change anything, or all of the suggestion.

schemas/public/ekscluster-kfd-v1alpha2.json
```diff
@@ -1581,10 +1595,18 @@
             }
           }
         },
+        "tsdbStartDate": {
+          "type": "string",
+          "format": "date",
+          "description": "The date loki have to switch to TSDB and schema v13"
```
ralgozino, line 1601: see comment above

schemas/public/kfddistribution-kfd-v1alpha2.json
```diff
@@ -598,10 +612,18 @@
             }
           }
         },
+        "tsdbStartDate": {
+          "type": "string",
+          "format": "date",
+          "description": "The date loki have to switch to TSDB and schema v13"
```
ralgozino, line 618: see comment above

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1182,10 +1196,18 @@
             }
           }
         },
+        "tsdbStartDate": {
+          "type": "string",
+          "format": "date",
+          "description": "The date loki have to switch to TSDB and schema v13"
```
ralgozino, line 1202: see comment above

defaults/ekscluster-kfd-v1alpha2.yaml
```diff
@@ -113,6 +113,8 @@ data:
           secretAccessKey: example
           accessKeyId: example
           bucketName: lokibucket
+        tsdbSchemav13Migration:
+          enabled: false
```
ralgozino, line 117: I don't see this being used in the templates

ralgozino (changes requested):
Hi! I left some suggestions on the code.

I also have a question: can we add a configurable parameter to set the logs retention now that we are using the TSDB? We could do that on another PR if you consider it out of scope of this one.

schemas/private/ekscluster-kfd-v1alpha2.json
lukeskywlkr: I think that we can introduce that, since the change if done incorrectly may result in loss of logs data, it will be very useful to have this in the autocomplete and on the KFD docs site. Thank you for pointing it out!

lukeskywlkr: shipped 📦

defaults/ekscluster-kfd-v1alpha2.yaml
lukeskywlkr: fixed 🩹

lukeskywlkr:
> Hi! I left some suggestions on the code.

Thank you for the suggestions! I have added them and it should be all in order now.

> I also have a question: can we add a configurable parameter to set the logs retention now that we are using the TSDB? We could do that on another PR if you consider it out of scope of this one.

Since @nutellinoit said it is out of scope maybe we can open an issue to track this request in order to add this feature in the next update? What do you think?

nutellinoit:
> Since @nutellinoit said it is out of scope maybe we can open an issue to track this request in order to add this feature in the next update? What do you think?

you can open an issue on the logging module to track this problem

ralgozino (approved):
Fine by me to create the issue for tracking the retention feature.

PR #334, opened 2025-01-14
fix policy template conditionals

### LINKED ISSUE #327: when using gatekeeper, if kyverno whitelist is set the manifests template will contain also patch to kyverno
**Describe the bug**
When using `spec.distribution.modules.policy.type: gatekeeper` , if `spec.distribution.modules.policy.kyverno.additionalExcludeNamespace` is set, the kustomize template will be generated with some patching to kyverno. This should not happen. The kustomize build will fail with accumulated errors.

**To Reproduce**
Set `furyctl.yaml` with:

```yaml
      policy:
        type: gatekeeper
        kyverno:
          additionalExcludedNamespaces: ["some-ns"]
```

**Expected behavior**
The generated template should not have any kyverno references

**Kubernetes (please complete the following information):**
 - Fury Distribution 1.30.0, all providers

ralgozino:
### Summary 💡

Don't patch Kyverno's config for excluding namespaces when the policy type is Gatekeeper.

Closes #327 

### Description 📝

Previously, if the furyctl.yaml included the configuration for excluding namespaces in Kyverno, the patch would be generated even if the policy.type was Gatekeeper.

Now, when the policy type is Gatekeeper, the patch to exclude namespaces in Kyverno won't be generated even if the configuration parameter is present in the furyctl.yaml file.

Other changes included:
- Simplified the patching logic of the `enforcementAction` for Gatekeeper's constraints using the `group` instead of patching each one individually, reducing the risk of forgetting to patch some constraint if a new one is added or not patching a new constraint that could be
added.
- Moved everything that is possible under the `patches` key in the kustomization file, being that `patchesStrategicMerge` has been deprecated in newer versions of Kustomize. Notice that due to a bug the patch for the infra nodes selection cannot be moved yet.
- Specified in the schemas docs that the `kyverno.validationFailureAction` value is for the **included** policies. Without the specification the user could understand that this would apply to all the policies when is not the case.
- Ran `make generate-go-models` and `make generate-docs`.

### Breaking Changes 💔

There are no breaking changes.

### Tests performed 🧪

> [!NOTE]
> All the following tests have been performed with KFD v1.31.0.

#### Tests before applying any fix

  - [x] Deployed KFD 1.31.0 with Policy type set to Kyverno and specified the configuration for both Kyverno and Gatekeeper in the furyctl.yaml file. -> deploy OK.
  - [x] Migrated `policy.type` from `kyverno` to `none` to swith later to Gatekeeper. -> FAILED
  - [x] Installed Gatekeeper with all the parameters configured and left all the configuration for Kyverno set too in the furyctl.yaml file. -> Replicated the error from the linked issue.

#### Tests after fixing the template

> [!NOTE]
> The tests were performed leaving the configuration for both Gatekeeper and Kyverno set in the furyctl.yaml file.
> i.e. when testing Gatekeeper, Kyverno parameters were also set and viceversa.

1. Gatekeeper tests:
  - [x] Deployed Gatekeeper with Kyverno still configured -> OK. Nothing from kyverno ends up on the rendered manifests.
  - [x] Policies configuration is applied, ns exemptions are applied, enforce method is applied.
  - [x] Test 1: Deny non compliant workload, and allow it in the excluded namespaces.
	```yaml
	installDefaultPolicies: true
	enforcementAction: deny
	additionalExcludedNamespaces:
	  - local-path-storage
	  - demo-ldap
	```
  - [x] Test 2: Warn non compliant workload, and not warn it in the excluded namespaces.
	```yaml
	installDefaultPolicies: true
	enforcementAction: warn
	additionalExcludedNamespaces:
	  - local-path-storage
	  - demo-ldap
	```
  - [x] Test 3: Don't install the included policies.
	```yaml
	installDefaultPolicies: false
	enforcementAction: warn
	# additionalExcludedNamespaces:
	#   - local-path-storage
	#   - demo-ldap
	```

  - [x] Test 4: Don't install included policies and don't fail even if excluded namespaces are set.
	```yaml
	installDefaultPolicies: false
	enforcementAction: warn
	additionalExcludedNamespaces:
	  - local-path-storage
	  - demo-ldap
	```

3. Kyverno tests:
  - [x] Policies configuration is applied, ns exemptions are applied, enforce method is applied.
  - [x] Test 1: Non compliant pods get blocked in all NS but not get blocked in the excluded namespaces
	```yaml
	validationFailureAction: Enforce
	installDefaultPolicies: true
	additionalExcludedNamespaces:
	  - local-path-storage
	  - demo-ldap
	```
  - [x] Test 2: Non compliant pods get scheduled.
	```yaml
	validationFailureAction: Audit
	installDefaultPolicies: true
	additionalExcludedNamespaces:
	  - local-path-storage
	  - demo-ldap
	```
  - [x] Test 3: NS are no longer excluded.
	```yaml
        validationFailureAction: Enforce
        installDefaultPolicies: true
        # additionalExcludedNamespaces:
        #   - local-path-storage
        #   - demo-ldap
	```

### Future work 🔧

While testing this fix I discovered that the migration for policy type from `kyverno` to `none` is not working, at least with KFD 1.31.0. I'll test other versions and create an issue with the details.


nutellinoit (approved):
LGTM! 🚀

ralgozino:
CI shows pending here on GitHub but it actually has passed on Drone. Merging.

PR #358, opened 2025-02-27
feat(kustomize): update to version v5 (5.6.0)

kriive:
### Summary 💡
Upgrade Kustomize from version 3.10 to the latest version (at the time of writing this is 5.6.0).

Relates: https://github.com/sighupio/product-management/issues/630

### Description 📝
Since we were falling a little behind with the Kustomize versions, we are now updating to version v5 of Kustomize.

### Breaking Changes 💔
`--load_restrictor` flag is now `--load-restrictor`.

### Tests performed 🧪
- [x] Built a cluster with the existing v1.31.0 fury-distribution, which uses kustomize version 3.10.0.
- [x] Manually built a new `out.yaml` with the new kustomize version.
  - [x] Diff between the "old" `out.yaml` and the newly generated `out.yaml`. Checked if there was any major difference.
- [x] Built a cluster with the not-yet existing fury-distribution with the new kustomize version 5.6.0.
- [x] Tested plugins apply with the new kustomize version (5.6.0).
- [x] Tested custom patches with the new kustomize version (5.6.0). 

### Future work 🔧

Follow implementation on `furyctl` too.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -16,7 +16,7 @@ The distribution is maintained with ❤️ by the team [SIGHUP](https://sighup.i
 
 ## Breaking changes 💔
 
-- TBD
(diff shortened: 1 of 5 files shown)


ralgozino (commented):
We should mention this change in the release notes (unreleased.md file) because it has impact on the plugins. Users may need to adjust their plugins if are using some kustomize 3 feature that has been deprecated.

ralgozino (changes requested):
as per previous comment

nutellinoit (changes requested):
I saw that the major tests were executed, I only have one minor complain, overall LGTM

ralgozino (approved):

nutellinoit (approved):
LGTM

PR #359, opened 2025-03-04
Add etcd backup over S3 and PVC 

kriive:
### Summary 💡

This PR adds a snapshot solution for the ETCD database.

Closes: https://github.com/sighupio/product-management/issues/82

Relates: https://github.com/sighupio/product-management/issues/82

### Description 📝

This PR leverages the snapshot functionality of `etcdctl` and saves it to a remote location, like an Object Storage such as Minio or locally, into a Persistent Volume Claim.

### Breaking Changes 💔

Hopefully nothing. 🙈 

### Tests performed 🧪

- [x] Tested the change with `etcdBackup.backupPrefix` set to "" and `etcdBackup.type` set to "s3", with credentials for Minio. Expected result: backup present on Minio, with no prefix in the name, only `<date>.etcdb`. Actual result: same as expected.
- [x] Tested the change with the same cluster as the previous test case, but in this case `etcdBackup.type` was set to "none". Expected result: the previously present _CronJob_ and its _ConfigMap_ and _Secret_ shouldn't be present anymore. Actual result: same as expected.
- [x] Tested the change with the same cluster as the previous test case, but in this case `etcdBackup.type` was set to "pvc", with an already created PersistentVolumeClaim. Expected result: backup present on the PVC. Actual result: same as expected.
- [x] Tested the change with the same cluster as the previous test case, but in this case `etcdBackup.type` was set to "all", with an already created PersistentVolumeClaim. Expected result: backup present on both the PVC and the Minio bucket. Actual result: same as expected.

### Future work 🔧

Support other providers beside local PVC and S3.


schemas/public/onpremises-kfd-v1alpha2.json
nutellinoit, line 1: There are too many changes on this file, why the overall linting has changed so much?

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1835,10 +1662,64 @@
...
+              "type": "string",
+              "description": "The access key ID (username) for the external S3-compatible bucket."
+            },
+            "bucketName": {
+              "type": "string",
+              "description": "The bucket name of the external S3-compatible object storage."
+            },
+            "retentionTime": {
+              "type": "string",
+              "description": "The retention time of the external S3-compatible object storage."
```
nutellinoit, line 1708: add an example string to the description

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1835,10 +1662,64 @@
...
+          "description": "The cron expression for the `full` backup schedule (default `0 1 * * *`)."
+        },
+        "s3": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Configuration for the remote S3 backend, where Etcd backup-restore will (optionally) send its backups.",
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "description": "If true, Etcd backup will save snapshots on S3."
```
nutellinoit, line 1684: If not true, what happens?

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +45,33 @@ patchesStrategicMerge:
...
+configMapGenerator:
+  - name: etcd-backup-config
+    behavior: replace
+    literals:
+      - target={{ print "minio:" .spec.distribution.modules.dr.etcdBackup.s3.bucketName }}
+      - retention={{ .spec.distribution.modules.dr.etcdBackup.s3.retentionTime }}
+{{- end }}
+{{- end }}
+
 secretGenerator:
```
nutellinoit, line 63: Are we sure this will not break aws provider? It's a kustomize statement without array elements in that case, If I understood correctly all the if opening and closing

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+{{- if and (index .spec.distribution.modules.dr "etcdBackup") (.spec.distribution.modules.dr.etcdBackup.s3)}}
+[minio]
+type = s3
+provider = Minio
```
nutellinoit, line 8: what if we want to use a real S3 bucket?

nutellinoit (changes requested):
I left some comments 👇

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
kriive: The Minio protocol should be the same as the S3 one, so I expect nothing would change. I can test with a real S3 bucket and see if something breaks.

schemas/public/onpremises-kfd-v1alpha2.json
kriive: I regenerated the Go models. I don't know about differences in linting, this file is autogenerated.

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: I figured that we were doing this in other places, so I though it would be fine: https://github.com/sighupio/fury-distribution/blob/97c89b5a4e1b4bef28e586234092329bf0c206b6/templates/distribution/manifests/monitoring/kustomization.yaml.tpl#L123

schemas/public/onpremises-kfd-v1alpha2.json
kriive: It won't enable any backup, at the moment.

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: Check this with EKS.

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
kriive: Add support for `insecure` use consistency and check if AWS S3 works out-of-the-box and consistently with what we're doing with Velero.

schemas/public/onpremises-kfd-v1alpha2.json
kriive: This was changed to use the `type` field, which can be `pvc`, `s3`, `all` or `none`.

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: Tested with `none` and it doesn't break.

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
kriive: Added support for `insecure`. Behaviour is consistent with other places, like Velero. AWS S3 not checked yet.

schemas/public/onpremises-kfd-v1alpha2.json
kriive: Resolved this, `Zed` decided it didn't like the current linting. Maybe in the future we can add a formatter for JSON Schema in the `Makefile`.

kriive:
Now ready for review. The only issue that I saw is if I replace (i.e. reapply the `furyctl.yaml`) the CronJob while a Job is running, one of the containers can error out because they can't find the old replaced ConfigMap. This is due to the `restart: OnError` policy.

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -32,6 +37,21 @@ resources:
   - resources/eks-velero-volumesnapshotlocation.yml
 {{- end }}
 
+{{- if eq .spec.distribution.common.provider.type "none" }}
```
ralgozino, line 40: IMPORTANT: notice that this is also true for the `KFDDistribution` kind.

I guess we can let the user enable etcdBackup with the KFDDistribution kind, having in mind that the (not managed by furyctl) underlying cluster must be made with kubeadm. If we want to do this, **you will need to update the schema for KFDDistribution to include the relevant fields too**.

There are several places where this if condition is repeated.

defaults/onpremises-kfd-v1alpha2.yaml
```diff
@@ -230,6 +230,21 @@ data:
         nodeSelector: null
         tolerations: null
       # the standard configuration for velero on the dr module
+      etcdBackup:
+        type: "none"
+        persistentVolumeClaim:
+          schedule: "0 1 * * *"
+          claimName: test
```
ralgozino, line 237: `claimName` should not be a default, make the user set it

defaults/onpremises-kfd-v1alpha2.yaml
```diff
@@ -230,6 +230,21 @@ data:
...
+          retentionTime: 10d
+        s3:
+          schedule: "0 1 * * *"
+          endpoint: s3.eu-west-1.amazonaws.com
+          insecure: false
+          accessKeyId: example
+          secretAccessKey: example
```
ralgozino, line 244: these values should not be set by default

defaults/onpremises-kfd-v1alpha2.yaml
```diff
@@ -230,6 +230,21 @@ data:
...
+          retentionTime: 10m
+          bucketName: etcdbucket
```
ralgozino, line 246: this value should not be set by default

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1874,10 +1877,87 @@
...
+              "description": "The bucket name of the external S3-compatible object storage."
+            },
+            "retentionTime": {
+              "type": "string",
+              "description": "The retention time of the external S3-compatible object storage. Follows rclone's `min-age` format. Example: '30d' for 30 days."
+            }
+          }
+        }
+      }
+    },
```
ralgozino, line 1960: we are missing some conditions here. We should make the schema validate that:

- If etcdBackup.type is pvc, then the PersistenVolumeClaim object must be filled.
- If etcdBackup.type is s3, then the s3 object must be filled.
- If etcdBackup is all, then both PersistentVolumeClaim and s3 must be filled.

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1874,10 +1877,87 @@
...
+              "type": "string",
+              "description": "The PersistentVolumeClaim name where the backup will be saved."
+            },
+            "retentionTime": {
+              "type": "string",
+              "description": "The retention time of the backups inside the PersistentVolumeClaim. Follows rclone's `min-age` format. Example: '30d' for 30 days."
+            }
+          }
+        },
+        "s3": {
```
ralgozino, line 1924: there's missing here the required properties definition, for a user to use the s3 type, they must set all fields but schedule, retentionTime and insecure that can be left to their default values or changed if wanted.

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1874,10 +1877,87 @@
...
+            "all"
+          ],
+          "description": "The type of the etcd backup to enable, options are:\n- `none`: no etcd backup CronJob will be installed and no etcd backup will be performed.\n- `s3`: the etcd-backup-s3 package will be enabled. It will deploy a CronJob which continuously snapshots a healthy etcd node and will save the backups in a configured S3 bucket.\n- `pvc`: the etcd-backup-pvc package will be enabled. It will deploy a CronJob which continuously snapshots a healthy etcd node and will save the backups in a configured PersistentVolumeClaim.\n- `all`: both kinds of backups will be enabled.\n\nDefault is `none`."
+        },
+        "backupPrefix": {
+          "type": "string",
+          "additionalProperties": false,
+          "description": "A prefix to be prepended to the backup filenames. If unset, the prefix defaults to the cluster's name."
+        },
+        "persistentVolumeClaim": {
```
ralgozino, line 1905: we are missing the required fields for persistentVolumeClaim, if a user wants to enable this, they must set the claimName, the rest of the fields can be left to their default value or specified if wanted.

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
   - patches/velero-schedule-manifests.yml
   - patches/velero-schedule-full.yml
 {{- end }}
+{{- if eq .spec.distribution.common.provider.type "none" }}
+{{- if ne .spec.distribution.modules.dr.etcdBackup.type "none" }}
+  - patches/etcd-backup-schedule.yml
+{{- end }}
+{{- end }}
```
ralgozino, line 68: you can (should) have this patch in the `patches` above too, simplifying a little the logic.

Notice that with kustomize v5 patchesStrategicMerge has been marked as deprecated and you should switch to patches anyway sooner o later.

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
...
+  - name: etcd-backup-s3-config
+    behavior: replace
+    literals:
+      - target={{ print "minio:" .spec.distribution.modules.dr.etcdBackup.s3.bucketName }}
+      - retention={{ .spec.distribution.modules.dr.etcdBackup.s3.retentionTime }}
+    {{- if eq (index .spec.distribution.modules.dr.etcdBackup "backupPrefix") nil }}
+      - backup-prefix={{ print .metadata.name "-" }}
+    {{- else}}
+      - backup-prefix={{ .spec.distribution.modules.dr.etcdBackup.backupPrefix }}
+    {{- end}}
```
ralgozino, line 83: You can just check that index is true, that makes it a little more legible, 
I HAVEN'T TESTED IT, but I'm pretty sure this works:

```suggestion
    {{- if index .spec.distribution.modules.dr.etcdBackup "backupPrefix" }}
      - backup-prefix={{ .spec.distribution.modules.dr.etcdBackup.backupPrefix }}
    {{- else}}
      - backup-prefix={{ print .metadata.name "-" }}
    {{- end}}
```

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
...
+      - ETCDCTL_CACERT=/etcd/etcd/ca.crt
+      - ETCDCTL_CERT=/etcd/apiserver-etcd-client.crt
+      - ETCDCTL_KEY=/etcd/apiserver-etcd-client.key
+{{- end }}
+{{- if eq .spec.distribution.modules.dr.etcdBackup.type "all" "pvc" }}
+  - name: etcd-backup-pvc-config
+    behavior: replace
+    literals:
+      - retention={{ .spec.distribution.modules.dr.etcdBackup.persistentVolumeClaim.retentionTime }}
+    {{- if eq (index .spec.distribution.modules.dr.etcdBackup "backupPrefix") nil }}
```
ralgozino, line 96: same as above

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
...
+      - backup-prefix={{ print .metadata.name "-" }}
+    {{- else}}
+      - backup-prefix={{ .spec.distribution.modules.dr.etcdBackup.backupPrefix }}
+    {{- end}}
+  - name: etcd-backup-s3-certificates-location
+    behavior: replace
+    literals:
+      - ETCDCTL_CACERT=/etcd/etcd/ca.crt
+      - ETCDCTL_CERT=/etcd/apiserver-etcd-client.crt
+      - ETCDCTL_KEY=/etcd/apiserver-etcd-client.key
```
ralgozino, line 89: do you need to do this? aren't these the default values from the module?

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
...
+      - backup-prefix={{ print .metadata.name "-" }}
+    {{- else}}
+      - backup-prefix={{ .spec.distribution.modules.dr.etcdBackup.backupPrefix }}
+    {{- end}}
+  - name: etcd-backup-pvc-certificates-location
+    behavior: replace
+    literals:
+      - ETCDCTL_CACERT=/etcd/etcd/ca.crt
+      - ETCDCTL_CERT=/etcd/apiserver-etcd-client.crt
+      - ETCDCTL_KEY=/etcd/apiserver-etcd-client.key
```
ralgozino, line 106: same as above

templates/distribution/manifests/dr/kustomization.yaml.tpl
```diff
@@ -41,13 +61,65 @@ patchesStrategicMerge:
...
+{{- end }}
+{{- end }}
+{{- end }}
 
 {{- if eq .spec.distribution.common.provider.type "none" }}
-{{- if eq .spec.distribution.modules.dr.velero.backend "externalEndpoint" }}
 secretGenerator:
+{{- if eq .spec.distribution.modules.dr.velero.backend "externalEndpoint" }}
```
ralgozino, line 113: notice that with this change, you can end up with a `secretGenerator:` line that has nothing below. I think kustomize handles this gracefully, but double check it if you haven't.

templates/distribution/manifests/dr/patches/etcd-backup-schedule.yml.tpl
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+{{- if eq .spec.distribution.common.provider.type "none" }}
```
ralgozino, line 5: same as for the other file

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+{{- if eq .spec.distribution.modules.dr.etcdBackup.type "all" "s3" }}
+[minio]
```
ralgozino, line 6: let's call this `s3` instead of `minio`, the name of the target storage should generic, it could be that the user does not uses minio

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
ralgozino: I don't see minio in rclone's docs though: https://rclone.org/docs/#configure

Maybe it is an alias for S3?

Anyway, working with S3 is a must, don't forget to check this.

ralgozino (changes requested):

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: Yes, I specifically checked for `nil` because otherwise Go's template engine considers a nil value (i.e. a non existing field) the same as the zero value (falsy-values). 

The rationale here is if `backupPrefix`'s left unset (i.e. nil value) we apply the cluster's name, if empty we want an empty prefix (i.e. only the date + .etcdb).

ralgozino: makes sense

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: Yes, you're right.

schemas/public/onpremises-kfd-v1alpha2.json
```diff
@@ -1874,10 +1877,87 @@
...
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Configuration for the PersistentVolumeClaim where `etcd-backup-pvc` will store its backups.",
```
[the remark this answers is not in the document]
kriive: I would go with `pvc`, it is simple and it matches the `type`.

templates/distribution/manifests/dr/kustomization.yaml.tpl
kriive: I can add a comment to better explain the rationale!

ralgozino: that would be great

templates/distribution/manifests/dr/secrets/rclone.conf.tpl
kriive: It's [here](https://rclone.org/s3/#minio), but I agree with you, we should use the S3 main provider.

ralgozino (changes requested):
one minor last change to the schema.

I notice that if you set dr.type to `none` then etcdBackup won't be deployed. I think that etcdBackup and Velero should be independent from each other. I may not want to deploy velero because I'm using other software but I may still want to have etcd backups.

Exmaple furyctl.yaml config:

```
...
spec:
  distribution:
    modules:
      dr:
        type: none
        etcdBackup:
          type: pvc
          pvc:
            claimName: etcd-backup
```

Maybe we should add a new `etcdBackup` or `etcdBackupOnly` to `dr.type`?

thoughts @nutellinoit @kriive ?

defaults/onpremises-kfd-v1alpha2.yaml
```diff
@@ -229,22 +229,17 @@ data:
...
           retentionTime: 10d
         s3:
           schedule: "0 1 * * *"
-          endpoint: s3.eu-west-1.amazonaws.com
+          retentionTime: 10d
           insecure: false
-          accessKeyId: example
-          secretAccessKey: example
-          retentionTime: 10m
           bucketName: etcdbucket
```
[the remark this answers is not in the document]
kriive: Nice catch, yes that was a leftover.

kfd.yaml
ralgozino, line 1: commited by mistake? :D

kriive: lolllll

kriive: I knew this would happen

PR #362, opened 2025-03-12
fix: small fixes in template config files

speziato:
### Summary 💡

This PR fixes two small issues with the template config files that `furyctl create config` uses:

- Link to documentation in the heading: correction after the refactoring of the documentation website
- Keepalived passphrase: after recent changes in the schema, the `spec.kubernetes.loadBalances.keepalived.passphrase` parameter in the OnPremises template needed to be changed

### Description 📝

When you create a `furyctl.yaml` config file using `furyctl create config`, the heading is directing users to an old documentation link that does not work anymore, because the refactoring of the documentation site changed the path of the Providers' explanations.

Also, after the [recent changes to the schema](https://github.com/sighupio/fury-distribution/commit/8a647a748032d3b99f89fdc9be63a7f569ebb800), the template config file for the OnPremises provider failed validations during the creation with `furyctl create config` because the template contained a keepalived passphrase longer than the allowed max length. This PR aligns the template with the schema's new constraints.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the link after using `furyctl create config --kind {OnPremises|KFDDistribution|EKSCluster} --version v1.31.0 --distro-location <path-to-local-kfd-distro>`

### Future work 🔧

N/A


ralgozino (approved):
LGTM! nice catch

PR #357, opened 2025-02-27
feat: add e2e tests for eks installer

Filo01:
### Summary 💡

Added e2e tests for EKSInstaller
The CI tests are triggered by tags `e2e-full-**`
Everyday at 11PM aws nuke is ran to cleanup leftover resources in case of errors during the e2e pipeline

### Description 📝

- Added e2e tests for EKSInstaller in `tests/e2e/ekscluster` + `tests/e2e/ekscluster-upgrades`
- The CI tests are triggered by tags `e2e-full-**`
- The tests create an eks cluster and then try to apply some modifications to the installed modules
- After the tests are run the cluster is deleted immediately
- at 11PM every day the CI launches aws-nuke to delete all the resources that cost money on aws to cleanup test resources, it can also be triggered via a tag `aws-nuke-ci-**`

### Tests performed 🧪

- ran locally with drone cli
- ran on actual drone ci

### Future work 🔧

To avoid using a priviledged runner on the drone pipeline we decided against using the vpn, so there are no tests for it.

While developing the tests I encountered 2 small issues:
- You cannot use the aws region `us-east-1` for the terraform state bucket/velero bucket because it's the default when not specified on aws and aws rejects aws region `us-east-1` if explicitly set. This causes an error since it's mandatory in the furyctl.yaml schema -> there is already[ an open issue](https://github.com/sighupio/fury-distribution/issues/270)
- Taints may not be properly applied to the `mimir` module (I tried running the tests with a single `infra` nodegroup with taints and tolerations and mimir pods would remain unschedulable - I have not investigated further) -> I will try to reproduce it and open an issue
- I added a temporary fix in the e2e test because pomerium would not delete the dns entry and it would break  -> I will try to reproduce it and open an issue

PR #364, opened 2025-03-18
fix(templates/distribution/monitoring/minio): use right node placement

### LINKED ISSUE #363: monitoring/minio: missing nodeSelector and Tolerations in job
### Describe the bug

The minio `minio-monitoring-buckets-setup` Job is missing the common nodeSelector and  Tolerations, so it won't get scheduled if you only have infra nodes. This breaks the deployment of the distribution phase because kapp keeps waiting for the job to complete.

### To Reproduce

Create a cluster with only tainted nodes and mimir as monitoring type. Add the toleration for the taint to `spec.distribution.common`.

Apply the configuration and check that the distribution phase gets stuck because is waiting for the job to finish but it is in pending state.

### Expected behavior

The minio monitoring setup job should use the same selector and tolerations defined in common.

### Software versions

 - Kubernetes Fury Distribution version: v1.31.0
 - furyctl version: 0.31.1

ralgozino:
### Summary 💡

Use the right node placement for Monitoring's Minio:
- Add missing placement patch to Minio's setup job.
- Use Minio's overrides instead of Mimir's for Minio's StatefulSet placement.

Closes #363 

Relates: #357 

### Description 📝

See Summary and Linked Issue.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested that Minio gets placed in the right nodes using the `.spec.distribution.common` tolerations and nodeSelectors.
- [x] Tested that Minio gets placed in the right nodes using the `.spec.distribution.modules.monitoring.minio.overrides` nodeSelector and tolerations.

### Future work 🔧

None

### DIFF
--- a/templates/distribution/manifests/monitoring/patches/infra-nodes.yml.tpl
+++ b/templates/distribution/manifests/monitoring/patches/infra-nodes.yml.tpl
@@ -173,6 +173,7 @@ spec:
       tolerations:
         {{ template "tolerations" $mimirArgs }}
 {{- if eq .spec.distribution.modules.monitoring.mimir.backend "minio" }}
+{{- $minioArgs := dict "module" "monitoring" "package" "minio" "spec" .spec }}
 ---
 apiVersion: apps/v1
 kind: StatefulSet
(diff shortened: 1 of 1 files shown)


stefanoghinelli (approved):
LGTM!

PR #371, opened 2025-03-26
Feat/upgrade logging module v5.0.0

lukeskywlkr:
### Summary 💡

This PR adds the changes needed to migrate to Loki version 3.4.2.

Relates:
https://github.com/sighupio/module-logging/pull/181
https://github.com/sighupio/module-logging/pull/183

### Description 📝

- Renaming secrets and adding `minio-credentials-loki` in `templates/distribution/manifests/logging/kustomization.yaml.tpl`
- Change resource type and add new components (index-gateway, query-frontend) in `templates/distribution/manifests/logging/patches/infra-nodes.yml.tpl` and `templates/distribution/manifests/logging/patches/loki.yml.tpl` (in the latter the querier was missing and it has been added)
- Use new Loki config in `templates/distribution/manifests/logging/patches/loki-config.yaml.tpl`
- Add minio credentials template for Loki
- Minio/external storage credentials have been parametrized in the new configuration

### Breaking Changes 💔

- The new Loki configuration template is not compatible with the previous one
- Minio/external storage credentials have been parametrized and the new format is not compatible with the previous configuration.

### Tests performed 🧪

- [X] Tested the change with KFD version 1.31.0
- [X] Tested an upgrade from the previous version 2.9.10 (Logging Core Module 4.0.0)
- [X] Tested with a clean install on kind with Kubernetes v1.32.2
- [X] Tested with an upgrade with`furyctl` from KFD v1.31.0 to this version - K8S v1.32.2

PR #386, opened 2025-04-08
docs(schemas): update description for loki's tsdbDate field

ralgozino:
### Summary 💡

Update documentation for the `tsdbStartDate` field in the Loki module to clarify that the change from BoltDB to TSDB and schema v11 to v13.

Also, correct the wording to indicate that the field is immutable once set and should be set to the past before upgrading to certain versions.

Fix the mentioned version for 1.30.x

NOTE: this PR includes `make generate-go-models` and `make generate-docs`.

Relates: #371 

### Description 📝

As per summary

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Checked that the docs render properly on markdown.

### Future work 🔧

- Maybe we should set a default for the `tsdbStartDate` field, now that the migration is done.
- Port these changes to the docs site.

### DIFF
--- a/docs/schemas/ekscluster-kfd-v1alpha2.md
+++ b/docs/schemas/ekscluster-kfd-v1alpha2.md
@@ -2618,11 +2618,11 @@ The memory request for the Pod. Example: `500M`.
 
 ### Description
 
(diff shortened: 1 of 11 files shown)


lukeskywlkr (approved):
LGTM! 🚀

stefanoghinelli (approved):
LGTM

Issue #360, opened 2025-03-06
`immutable` behaviour in KFD


kriive:
The `immutable` keyword is used inside the `rules/*-kfd-v1alpha2.yaml` files to highlight which fields of the `furyctl.yaml` file are not to be changed between two runs of `furyctl apply`.
 
These immutability rules are particularly useful in cases when you have fields describing resources which cannot be easily mutated or changed, such as fundamental blocks of the cluster, like the Pod and Service CIDRs.
 
- Code only checks the single YAML path, i.e. if we have an `immutable` rule on path `.spec.kubernetes.foo` and we change the child key `.spec.kubernetes.foo.bar` and we run`furyctl apply`, `furyctl` shows no errors.
- The behaviour is defined [here](https://github.com/sighupio/furyctl/blob/main/pkg/diffs/checker.go#L157-L168). 
- If we want to make not only the *parent* key but also all its children immutable, we need to swap `==` with  `strings.HasPrefix`. Since this change would impact quite a bit of pieces, we might break some implicit assumptions so we must check thoroughly that we don't end up breaking stuff.

PR #387, opened 2025-04-11
Fix immutability on control planes

kriive:
### Summary 💡
This PR addresses a bug preventing control planes nodes array to be marked as immutable under the OnPremises provider. Control plane nodes were set as immutable in the 1.31.1 release cycle because we don't have any support to scale the etcd cluster with the number of control plane nodes. Moreover, the etcd cluster can potentially be placed on control plane nodes or set-up on separate machines, adding further cases to the etcd scaling support, which at the moment we don't support.

Closes: #360 

Refers:  https://github.com/sighupio/product-management/issues/651

### Description 📝
Immutable behaviour in the OnPremises provider is checked during the preflight phase, after having checked the availability of the hosts set in the `.spec.kubernetes.masters.hosts.*` keys in `furyctl.yaml`. The availability of the hosts is performed with an Ansible Playbook, called `verify-playbook.yaml`.

This playbook is used to both check the availability of the hosts and to get the kubeconfig necessary to talk with the cluster. It does so by accessing the `/etc/kubernetes/admin.conf` file from all hosts inside the `.spec.kubernetes.masters.hosts.*` key and copies the `admin.conf` in a local (to the machine running `furyctl`) directory. The surrounding logic about this playbook is the following: if, for any reason, the playbook fails, we mark the cluster as "pristine" and we proceed to install the cluster as if it's yet to be installed, bypassing the immutability checks, since this cluster is now treated as brand new.

So, if we installed the cluster the first time with `furyctl apply` and then proceeded to edit the `furyctl.yaml` adding a new master, and re-ran `furyctl apply`, `verify-playbook.yaml` fails (since the `master` hosts are now two, and one of them doesn't contain the `admin.conf` triggering a failure), marking the already installed cluster as "brand new", bypassing the immutability checks and triggering all the new installation ceremonies.

The fix is simple, and is to make it fail if and only if no host contains a `/etc/kubernetes/admin.conf` file. This makes the check works as intended and lets the immutability checks do their intended job.

### Breaking Changes 💔
Nothing.

### Tests performed 🧪
 - [x] Installed a cluster with only one control plane set, edited its `furyctl.yaml` file and retried `furyctl apply`. Expected output: *immutability check failed*. Actual result: as expected.

### Future work 🔧
Introduce etcd cluster scaling so that we can drop the immutability.

### DIFF
--- a/templates/preflight/onpremises/verify-playbook.yaml
+++ b/templates/preflight/onpremises/verify-playbook.yaml
@@ -11,9 +11,17 @@
         path: /etc/kubernetes/admin.conf
       register: admin_conf_stat
 
+    - name: Getting admin.conf kubeconfig
+      become: yes
+      fetch:
+        src: /etc/kubernetes/admin.conf
+        dest: "{{ kubernetes_kubeconfig_path }}/admin.conf"
+        flat: yes
+      when: admin_conf_stat.stat.exists
+            
     - name: Set fact if admin.conf exists
       set_fact:
-        admin_conf_exists: "{{"{{ admin_conf_stat.stat.exists }}"}}"
+        admin_conf_exists: "{{ admin_conf_stat.stat.exists }}"
       when: admin_conf_stat.stat.exists
 
 - name: Aggregate results and fetch admin.conf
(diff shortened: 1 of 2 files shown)


ralgozino (approved):
LGTM!

Filo01 (approved):

PR #392, opened 2025-04-17
Add secured images to e2e

kriive:
### Summary 💡

This PR adds support for testing images present in the fury-secured registry.

Relates: https://github.com/sighupio/product-management/issues/528

### Description 📝
Add a new test case that uses the fury-secured registry. The idea here is to re-deploy the cluster with all the modules enabled, using the hardened registry and then reset it to a vanilla cluster, dropping the custom registry and going back to the default one.

This is done by applying a slightly edited `init` KFD cluster, similar to the one of the first test, with the only difference being the registry changed to the secured one.

### Breaking Changes 💔
Nothing.

### Tests performed 🧪
- [x] Run tests locally with the new manifest with the custom secured registry. 

### Future work 🔧
Nothing.


ralgozino (commented):
Can you please add to the PR description a some details on what you've done?

From the code diff it seems that you are redeploying to the same cluster with the patched images. But I'm not sure that is the intended behaviour, being that you have made a separated "init" furyctl.yaml file.

ralgozino (changes requested):
as discussed, let's use an in-place command to set the registry so we don't have to maintain yet another furyctl.yaml file

ralgozino (approved):
lgtm

Issue #365, opened 2025-03-19
Migrating Ingress to `none` with network policies enabled results in error


ralgozino:
### Describe the bug
In a cluster with the beta feature `networkPoliciesEnabled` set to `true`, switching ingress type `dual` or `single` to `none` when `ingress.tls.provider` is `secret` results in an error:

```
ERRO error while creating cluster: error while executing distribution phase: error while executing phase: error running core distribution phase: error applying manifests: error while running shell: /bin/sh sh <redacted>/.furyctl/multipass/distribution/scripts/apply.sh: command failed - exit status 1
out:
err: Error: accumulating resources: accumulation err='accumulating resources from 'ingress': read <redacted>/.furyctl/multipass/distribution/manifests/ingress: is a directory': recursed accumulation of path '<redacted>/.furyctl/multipass/distribution/manifests/ingress': accumulating resources: accumulation err='accumulating resources from 'policies': read <redacted>/.furyctl/multipass/distribution/manifests/ingress/policies: is a directory': couldn't make target for path '<redacted>/.furyctl/multipass/distribution/manifests/ingress/policies': kustomization.yaml is empty
``` 

### To Reproduce

- Create an on-prem cluster with `.spec.distribution.common.networkPoliciesEnabled: true` and `.spec.distribution.modules.ingress.nginx.type` set single (or dual) **and `tls.provider: secret`**.
- Apply the configuration.
- Change `networkPoliciesEnabled` to `false`.
- Reapply the configuration and see the error above.

### Expected behavior

Network policies should be disabled without error.

### Software versions

 - Kubernetes Fury Distribution version: 1.31.0
 - furyctl version: 0.31.0

ralgozino:
I did a little more debugging, the problem is not in the migration, it is present also when deploying a cluster from scratch with:
1. `spec.distribution.modules.ingress.type: none`
2. `spec.distribution.modules.ingress.tls.provider: secret`
3. `spec.distribution.common.networkPoliciesEnabled: true`

The problem is also present if you start the cluster with ingress type none, tls provider secret, and network policies disabled and try to enabled them afterwards. 

The issue seems to be in the template of the kustomization.yaml file for the ingress network policies that does not consider the case when ingress type is none and tls provider is `secret`, generating an invalid file.

PR #393, opened 2025-04-18
fix(templates): add condition for network policies in Ingress module

stefanoghinelli:
### Summary 💡

With this PR network policies for the Ingress module are included in the root `kustomization.yaml` only when at least one condition between `spec.distribution.modules.ingress.nginx.tls.provider: certManager` or `spec.distribution.modules.ingress.nginx.type != none` is met and `spec.distribution.common.networkPoliciesEnabled` is true.

Closes: [#365](https://github.com/sighupio/distribution/issues/365)

### Description 📝

With the migration from Kustomize v3 to v5, the behavior around empty `kustomization.yaml` files has changed.
In Kustomize v3, a file with an empty resources list was silently ignored, in Kustomize v5 the same file causes an error during the build.

To fix this, we now include the network policies only when network policies are enabled and at least one of the above conditions applies.

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Tested the change with `spec.distribution.modules.ingress.nginx.tls.provider: certManager`
- [x] Tested the change with `spec.distribution.modules.ingress.nginx.tls.provider: secret`
- [x] Tested the change with `spec.distribution.modules.ingress.nginx.type: single`
- [x] Tested the change with `spec.distribution.modules.ingress.nginx.type: none`

### Future work 🔧

None.

### DIFF
--- a/templates/distribution/manifests/ingress/kustomization.yaml.tpl
+++ b/templates/distribution/manifests/ingress/kustomization.yaml.tpl
@@ -24,7 +24,7 @@ resources:
 
 {{- end }}
 
-{{ if eq .spec.distribution.common.networkPoliciesEnabled true }}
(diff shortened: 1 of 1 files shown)


ralgozino (approved):
lgtm

Issue #400, opened 2025-04-29
Changing logging type from loki to none fails when monitoring is disabled


ralgozino:
### Describe the bug

When trying to change logging.type from loki to none, the resources deletion step (pre-apply.sh script) fails because it tries to delete some objects using the non-available monitoring.coreos.com/v1 API.

### To Reproduce
1. Create a cluster with logging type `loki` and monitoring type `none`
2. Apply the configuration
3. Change the configuration of logging type from `loki` to `none`
4. See error similar to:

```
err: Error from server (NotFound): apiservices.apiregistration.k8s.io "v1.monitoring.coreos.com" not found
Error from server (NotFound): apiservices.apiregistration.k8s.io "v1.monitoring.coreos.com" not found
resource mapping not found for name: "opensearch-dashboards-service-monitor" namespace: "logging" from "delete-opensearch-dashboards.yaml": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
ensure CRDs are installed first
```

### Expected behavior

The logging resources deletion should not fail when monitoring is not present and I should be able to disable logging.

### Software versions
Please complete the following information:

 - SIGHUP Distribution version: 1.31.1
 - furyctl version: 0.32.1

## Additional context

1. The logic to exclude monitoring resources for the OpenSearch Dashboards package is missing here (the filtering is only being done for opensearch):

https://github.com/sighupio/distribution/blob/de9d02085edd086e9fcd740d3dd77622a172c47a/templates/distribution/scripts/pre-apply.sh.tpl#L47-L55

2. There's a copy & paste error on the `cp` copy, notice `tracing` instead of `logging`:

https://github.com/sighupio/distribution/blob/de9d02085edd086e9fcd740d3dd77622a172c47a/templates/distribution/scripts/pre-apply.sh.tpl#L98

Issue #398, opened 2025-04-24
Loki can't be enabled after a first installation of the cluster


ralgozino:
### Describe the bug

If a cluster has been created with logging type set to none, you can't set it to Loki afterwards because of the immutability of the `tsdbStartDate` field that does not let you change from `nil` to a date that you are required to set.

### To Reproduce

Steps to reproduce the behaviour:
1. Create a cluster with `logging.type: none`
2. Apply the cluster configuration
3. Change the `logging.type: loki` and add the `loki.tsdbStartDate` parameter as required by the schema.
4. See the following error:

```
ERRO error while creating cluster: error while executing preflight phase: error checking state diffs: immutable path changed: [immutable value changed: path .spec.distribution.modules.logging.loki.tsdbStartDate  oldValue <nil> newValue 2023-01-01]
```

### Expected behaviour

One should be able to enable Loki afterwards.

### Software versions

 - SIGHUP Distribution version: 1.31.1
 - furyctl version: 0.32.1

ralgozino:
Some more findings:

If in the config file used to create the cluster `logging.type` is `none` **but `logging.loki.tsdbStartDate` was set**, you don't face the error. This is because the value is saved in the `furyctl-config` secret so "it does not change" when you change the type to `loki` and does not trigger the immutability failure.

ralgozino:
fixed in v1.32.0 of the Distribution, for v1.31.1 and v1.30.2 the issue is still present and needs the workaround.

PR #401, opened 2025-04-30
fix(migrations/logging): to none without monitoring

ralgozino:
### Summary 💡

Fix migrations of logging from * to none when monitoring is disabled.

logging migrations to none (i.e. deleting logging resources) failed if _monitoring_ was none because we were trying to delete resources using the monitoring.coreos.com/v1 API that was not available. This fix filters out these type of resources from the deletion.

Closes: #400 

Relates: discovered while debugging #398 

### Description 📝

See summary

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the change with KFD version v1.31.1 and verified that the migration finishes successfully and all resources are deleted.

### Future work 🔧

None


stefanoghinelli (approved):
🚀

PR #403, opened 2025-05-06
fix(templates): ingress cert-manager solvers

### LINKED ISSUE #211: Furyctl validated correctly but `furyctl apply` reports missing values
## Scenario
I have a `furyctl.yaml` file for an OnPremises installation, which currently needs the CNI only:
```yaml
kind: OnPremises
metadata:
  name: fury-poc
spec:
  distributionVersion: v1.27.3
  ...
  distribution:
    modules:
      networking:
        type: "calico"
      ingress:
        baseDomain: internal.example.dev
        nginx:
          type: none
      logging:
        type: none
      monitoring:
        type: none
      ...
```
I made sure everything was okay by validating the config
```console
❯ furyctl validate config
INFO Downloading distribution...                  
INFO Compatibility patches applied for v1.27.3    
INFO configuration file validation succeeded      
INFO There is a newer release available: v0.28.0  
```

But then, when running the apply, I got the following error:
```console
❯ furyctl apply --phase distribution --skip-nodes-upgrade --force upgrades
INFO Downloading distribution...                  
INFO Compatibility patches applied for v1.27.3    
INFO Validating configuration file...             
INFO Downloading dependencies...                  
INFO There is a newer release available: v0.28.0  
INFO Validating dependencies...                   
INFO Running preflight checks                     
INFO Checking that the cluster is reachable...    
INFO Preflight checks completed successfully      
INFO Running preupgrade phase...                  
INFO Preupgrade phase completed successfully      
INFO Installing Kubernetes Fury Distribution...   
INFO Checking that the cluster is reachable...    
INFO Checking storage classes...                  
WARN No storage classes found in the cluster. logging module (if enabled), tracing module (if enabled), dr module (if enabled) and prometheus-operated package installation will be skipped. You need to install a StorageClass and re-run furyctl to install the missing components. 
ERRO error while creating cluster: error while executing distribution phase: error while executing phase: error preparing distribution phase: error copying from template: error generating from template files: error applying templates: error processing template: template: cert-manager-clusterissuer.yml.tpl:40:17: executing "cert-manager-clusterissuer.yml.tpl" at <.spec.distribution.modules.ingress.certManager.clusterIssuer.solvers>: map has no entry for key "solvers" filePath: /var/folders/mf/vblq1rpx01571403sbmyb6m40000gn/T/furyctl-3210227887/data/templates/distribution/manifests/ingress/resources/cert-manager-clusterissuer.yml.tpl 
```

I also tried setting the parameter `--force all` but the error persists.

ralgozino:
### Summary 💡

- Don't fail template rendering when cert-manager's solvers are not set.
- Always patch cert-manager's resources to use the nodeSelectors and Tolerations 

Fixes #211

### Description 📝

- Being that the default value for `ingress.tls.provider` is `certManager` without a `clusterIssuer.type` set, and there is not a default value for solvers, the template would fail trying to access the solvers key in the configuration. Changed to check if the key exists instead.

- When deploying the default configuration with `ingress.nginx.type: none` and `tls.provider` **not** set (so, cert-manager as per the defaults); the patch for adding `nodeSelectors` and `Tolerations` was not being applied. Now we apply always the patch, because cert-manager is always deployed.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested apply with `ingress.nginx.type: none` and no other fields specified -> template does not fail anymore and cert-manager gets deployed.
- [x] Tested apply with `ingress.nginx.type: none` and `tls.provider: none` ->  template does not fail anymore and cert-manager gets deployed.
- [x] Tested apply with `ingress.nginx.type: single` and `tls.provider: none` -> templates does not fail, both nginx and cert-manager get deployed.
- [x] Tested apply with `ingress.nginx.type: single` and `tls.provider: cert-manager` ->  templates does not fail, both nginx and cert-manager get deployed.
- [x] Tested apply with `ingress.nginx.type: single` and `tls.provider` not set ->  templates does not fail, both nginx and cert-manager get deployed.

### Future work 🔧

None

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -20,6 +20,7 @@ The distribution is maintained with ❤️ by the team [SIGHUP by ReeVo](https:/
 - [[#387](https://github.com/sighupio/distribution/pull/387)]: This PR fixed an issue that prevented the control planes nodes array to be treated as immutable under the OnPremises provider. The number of control plane nodes was originally set as immutable in the 1.31.1 release cycle because there isn't any support to scale the etcd cluster with the number of control plane nodes in the SIGHUP Distribution yet. The issue allowed users to change the number of the control plane, even if it was explicitly marked as immutable.
 - [[#393](https://github.com/sighupio/distribution/pull/393)]: This PR fixes an issue present when users start the cluster with ingress type none, TLS provider secret, and network policies disabled and try to enable them afterwards.
 - [[#401](https://github.com/sighupio/distribution/pull/401)]: Fixed an error in uninstalling the logging module that prevented switching from any logging type to `none` when the monitoring module was not installed.
+- [[#403](https://github.com/sighupio/distribution/pull/403)]: Fixed an error that prevented the generation of manifests when `ingress.nginx.type: none` and `ingress.certManager` configuration was not set.
 
 ### Security fixes
 
--- a/templates/distribution/manifests/ingress/kustomization.yaml.tpl
+++ b/templates/distribution/manifests/ingress/kustomization.yaml.tpl
@@ -54,9 +54,8 @@ patchesStrategicMerge:
 
   - patches/cert-manager-kapp-group.yml
 
-{{ if or (ne .spec.distribution.modules.ingress.nginx.tls.provider "none") (ne .spec.distribution.modules.ingress.nginx.type "none") }}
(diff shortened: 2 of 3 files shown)


stefanoghinelli (approved):
I did a few tests and it looks good :) lgtm

PR #409, opened 2025-05-15
fix: tsdbStartDate immutability when migrating from logging.type: none to logging.type: loki

Filo01:
### Summary 💡

fix immutability rules for `.spec.distribution.modules.logging.loki.tsdbStartDate`
it should be possible to migrate from `logging.type: none` to `logging.type: loki`

Closes: https://github.com/sighupio/distribution/issues/398

Relates:
- https://github.com/sighupio/furyctl/pull/610

### Description 📝

added new safe reducer to:
- `rules/ekscluster-kfd-v1alpha2.yaml`
- `rules/kfddistribution-kfd-v1alpha2.yaml`
- `rules/onpremises-kfd-v1alpha2.yaml`

### Breaking Changes 💔

none

### Tests performed 🧪

- [x] Tested with a kind cluster
  - migrating from logging.type: none to logging.type: loki -> does not get blocked
  - migrating from logging.type: loki to logging.type: opensearch -> does not get blocked
  - migrating from logging.type: loki to logging.type: none -> does not get blocked
  - migrating from logging.type: loki to logging.type: opensearch -> gets blocked, unsupported
  - regression test: migrating from ingress single to ingress dual -> still gets blocked

### Future work 🔧

none

### DIFF
--- a/rules/ekscluster-kfd-v1alpha2.yaml
+++ b/rules/ekscluster-kfd-v1alpha2.yaml
@@ -11,13 +11,6 @@ distribution:
     description: "changes to the logging module type have been detected. This will cause the reconfiguration or deletion of the current logging stack."
     safe:
       - from: none
-    unsupported:
-      - to: opensearch
-        from: loki
-        reason: "currently, switching from opensearch to loki is not supported. You need to first remove the current stack with type: none."
-      - to: loki
-        from: opensearch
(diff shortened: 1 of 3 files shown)


ralgozino (approved):
lgtm

now we can set a default to tsdbStartDate so it does not have to be set on new installations :tada:

PR #411, opened 2025-05-15
feat(logging): set default value for Loki's tsdbStartDate

ralgozino:
### Summary 💡

Since we have Loki v3 that does not support BoltDB anymore and TSDB is the default, we don't need to ask the user for a migration date anymore.

We set a default value for the tsdbStartDate in the past, so all the new logs will be written usign TSDB.

We had set the immutability rules for this field in #409

Relates:
- #409 

### Description 📝

As per summary

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Update from 1.31.1 with tsdbStartDate configured to 1.32.0 without tsdbStartDate configured. Expected output: immutability failure.
- [x] Update from 1.31.1 with tsdbStartDate configured to 1.32.0 with tsdbStartDate configured to a different date. Expected output: immutability failure.
- [x] Update from 1.31.1 with tsdbStartDate configured to 1.32.0 with tsdbStartDate configured to the same date. Expected output: apply successful.
- [x] Upgrade from 1.31.1 without tsdbStartDate set (logging `none`) to 1.32.0 with logging disabled. Expected output: apply successful.
- [x] Upgrade from 1.31.1 without tsdbStartDate set (logging `none`) to 1.32.0 with logging type `loki` and without tsdbStartDate specified. Expected output: apply successful.
- [x] Clean install 1.32.0 with loki enabled without specifying tsdbStartDate.
- [x] Clean install 1.32.0 without loki, then enable it.

### Future work 🔧

None

### DIFF
--- a/defaults/ekscluster-kfd-v1alpha2.yaml
+++ b/defaults/ekscluster-kfd-v1alpha2.yaml
@@ -106,6 +106,7 @@ data:
           username: minio
           password: minio123
       loki:
+        tsdbStartDate: "2025-05-15" # we needed this field when migrating from BoltDB to TSDB. Now TSDB is the default.
         backend: minio
         externalEndpoint:
           endpoint: s3.eu-west-1.amazonaws.com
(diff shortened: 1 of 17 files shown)


stefanoghinelli (approved):
LGTM

PR #428, opened 2025-08-13
feat(logging): allow to set Fluentd and Fluentbit's resouces

ralgozino:
### Summary 💡

- Add configuration fields to the schemas to allow setting resources to Logging Operator's Fluentd and Fluentbit.
- Add configuration field to set Logging Operator's Fluentd replica count.
- Adjust templates to use the new fields.
- Update logging's kustomization.yaml file to use `patches` and drop deprecated syntax. Make logic a little clearer (hopefully)
- Update related documentation.

Relates:
- https://github.com/sighupio/module-logging/pull/191 

### Description 📝

We've seen lately that for production clusters the defaults were not enough and patching the resources requests and specially limits for the loggging stack components was a common use case.

This PR adds missing parameters to configure the logging stack from the furyctl.yaml file, under the `spec.distribution.modules.logging.operator` key, one for fluentd and one for fluentbit.

Example:

```yaml
spec:
  distritbution:
    modules:
      logging:
        type: loki
        operator:
          fluentd:
            replicas: 1
            resources:
              limits:
                cpu: "10000m"
          fluentbit:
            resources:
              requests:
                memory: 1M
...
```

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested using `main` version of the distro and pointing to the `feat/increase-default-resources` branch of the Logging module (that includes the new default values).
- [x] Tested NOT setting any of the new parameters and verified that the templates still work as expected and there are no errors of undefined keys, for example.
- [x] Tested with type Loki and setting fluentd replicas, requests and limits and fluentbit request and limits and several combinations of these options.
- [x] Tested with type OpenSearch and setting fluentd replicas, requests and limits and fluentbit request and limits and several combinations of these options.

### Future work 🔧

I considered adding also parameters to patch the **flush thread count** for the Outputs, but I decided to leave it for future PRs because:

1. We've bumped the default values (see related PR) for this parameter in the logging module
2. Configuration would became a little messy because we would need to add a key for each output, another option would be having a common value for all of the (cluster)Outputs.

IMO we can wait to see if the new default values cover most of the cases and then decide if we still need to expose that parameter.

### DIFF
--- a/defaults/ekscluster-kfd-v1alpha2.yaml
+++ b/defaults/ekscluster-kfd-v1alpha2.yaml
@@ -92,6 +92,9 @@ data:
             disableAuth: false
             host: ""
             ingressClass: ""
+      operator:
+        fluentd: {}
+        fluentbit: {}
       # can be none, opensearch, loki or customOutputs
       type: opensearch
       opensearch:
--- a/defaults/kfddistribution-kfd-v1alpha2.yaml
+++ b/defaults/kfddistribution-kfd-v1alpha2.yaml
@@ -85,6 +85,9 @@ data:
             disableAuth: false
             host: ""
             ingressClass: ""
+      operator:
+        fluentd: {}
+        fluentbit: {}
       # can be none, opensearch, loki or customOutputs
       type: opensearch
       opensearch:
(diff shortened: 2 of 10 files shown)


stefanoghinelli (approved):
Makes sense to hold off and evaluate how the default settings perform. Ship it!

PR #436, opened 2025-09-16
Fix Tigera podCidr in KFDDistribution

kriive:
### Summary 💡
This PR fixes the handling of `podCidr` when left unset in KFDDistribution.

### Description 📝
Previously, leaving `podCidr` empty in the KFDDistribution provider resulted in a `null` value being set in Tigera's `Installation` CR. This value is invalid and caused Calico installation to fail.

We're now avoiding the fallback to `.spec.kubernetes.podCidr` when the user is installing a cluster using the KFDDistribution provider, we just omit the `.calicoNetwork.ipPools` in the rendered template.

When `.calicoNetwork.ipPools` is left unset in the `Installation` CR, the Tigera Operator applies its defaulting logic:
1. On OpenShift, it reads the `Network` configuration and merges the settings into the `Installation`.
2. On kubeadm-based clusters, it reads the `kubeadm-config` ConfigMap.
3. On EKS, it creates a default IPPool with CIDR `172.16.0.0/16` and VXLAN encapsulation.
4. Otherwise, it creates a default IPPool with CIDR `192.168.0.0/16`.

### Breaking Changes 💔
None. The only change is that in the KFDDistribution provider we no longer fall back to `.spec.kubernetes.podCidr`; instead, we rely on the Tigera Operator’s defaulting logic.

### Tests performed 🧪
End-to-end tests and `furyctl dump template` for both KFDDistribution and OnPremises provider.

### Future work 🔧
None.

### DIFF
--- a/templates/distribution/manifests/networking/patches/infra-nodes-and-mask-tigera.yaml.tpl
+++ b/templates/distribution/manifests/networking/patches/infra-nodes-and-mask-tigera.yaml.tpl
@@ -23,8 +23,8 @@ spec:
           tolerations:
             {{ template "tolerations" ( merge (dict "indent" 12) $tigeraOperatorArgs ) }}
   {{- if ne .spec.distribution.common.provider.type "eks" }}
-  {{/* .spec.kubernetes.podCidr is filled only inside OnPremises clusters */}}
-  {{- if and (index .spec "kubernetes") (index .spec.kubernetes "podCidr") }}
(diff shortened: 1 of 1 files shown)


kriive:
I agree on setting the `podCidr` field required if `blockSize` is set, but I would keep it the same for both `KFDDistribution` and for `OnPremises` for consistency.

PR #439, opened 2025-09-17
fix: keep Loki replicas

### LINKED ISSUE #438: Applying the distribution phase resets replicas values for Loki components
### Describe the bug

Every time you apply the `distribution` phase, the replicas value for Loki components gets reseted to `1`, overwriting the value set by the HPAs.

This is a problem because:

1. If the HPA has raised the value it means that the scale up was needed to keep up with the workload, so setting it back to 1 probably creates disruption.
2. The apply of the distribution phase takes a long time beacause Loki components like the ingestor try to flush to the remote storage while going in graceful shutdown.

### To Reproduce
1. Deploy a cluster with Loki as logging type
2. Create some load so you see the HPA scale up some components like the Ingestor.
3. Apply the distribution phase, even without changes.
4. See in the debug logs that kapp is reseting the replicas value to 1.

### Expected behavior

Applying the distribution phase should not override the HPA-managed values.

### Software versions

 - SIGHUP Distribution version: v1.32.1 (but applies to all the versions)
 - furyctl version: 0.32.5

ralgozino:
### Summary 💡

Add kapp configuration definitions to avoid overriding the replicas number set by the HPAs we deploy together with Loki.

Fixes #438

### Description 📝

I took the chance and also added an exception to a field of the Logging Operator CRD that was being removed by kapp on every apply. This did not have the impact that changing the replicas for Loki's components has, but I fixed it while I was at it.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the change with SD version 1.33 (wip), verified that replica number is not overwritten.
- [x] Tested that there are no errors on a clean deploy.

### Future work 🔧

There are other places where we could do the same probably.

### DIFF
--- a/templates/distribution/manifests/logging/kapp-configs/logging-operator-crd.yaml
+++ b/templates/distribution/manifests/logging/kapp-configs/logging-operator-crd.yaml
@@ -1,0 +1,19 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# This file is used to tell kapp to ignore the changes in the conversion strategy
+# of the Logging Operator CRD.
+
+apiVersion: kapp.k14s.io/v1alpha1
+kind: Config
+metadata:
+  name: dummy-logging-operator-crds # needed so kustomize won't complain
+rebaseRules:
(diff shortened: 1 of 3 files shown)


stefanoghinelli (approved):
lgtm

PR #441, opened 2025-09-17
fix(templates/monitoring): make kustomization.yaml v5 compatible

ralgozino:
### Summary 💡

Update kustomization.yaml template file for the monitoring module to use Kustomize v5 syntax and drop deprecated features usage like PatchesStrategicMerge.

### Description 📝

See summary

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested that the apply with the monitoring module enable is still working.
- [x] Tested changing some options like `installEnhancedHPAMetrics` and `installDefaultRules` and verified that the patches are correctly applied.
-->

### Future work 🔧

None

### DIFF
--- a/templates/distribution/manifests/monitoring/kustomization.yaml.tpl
+++ b/templates/distribution/manifests/monitoring/kustomization.yaml.tpl
@@ -60,68 +60,68 @@ resources:
   - policies
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


stefanoghinelli (approved):
LGTM!

PR #447, opened 2025-09-24
fix: eks provider following alinux2 deprecation

kriive:
### Summary 💡

This PR fixes a number of issues related to EKS e2e tests.

### Description 📝
We deprecated the alinux2 image, so we rewrote some e2e tests in order to use alinux2023. Then we found an issue about AWS Load Balancer Controller, which was failing because of the new IMDSv2 required and enforced in alinux2023. This required us to introduce a sane default to `metadata_http_put_response_hop_limit`, which is now set to 2, in both eks-managed and self-managed nodes.

### Breaking Changes 💔
Users are required to migrate their `alinux2` node pools to `alinux2023` before attempting a v1.33 upgrade. 

### Tests performed 🧪
- Migrate `alinux2` to `alinux2023` and upgrade in both eks-managed and self-managed nodePools
- Installed the distro at 1.33 

### Future work 🔧
Work on tech debt: for example, Terraform provider is ancient.


kriive:
e2e tests are failing on CI: race conditions on the applications of some (apparently) unrelated CRDs. Manual testing (running the same pipeline on my local machine, with `drone exec`) results in a successful run.

PR #459, opened 2025-10-08
feat(on-prem): support kube-proxy-less clusters

ralgozino:
### Summary 💡

Add support for CREATING on-premises clusters without the kube-proxy.

Relates:
- https://github.com/sighupio/installer-on-premises/pull/137

>[!WARNING]
> This PR requires Tigera Operator >= v1.40.0 and the Networking module is currently including v1.38.6.
> Support for Tigera Operator >= v1.40.0 is being introduced in this PR:
> - https://github.com/sighupio/module-networking/pull/102

>[!WARNING]
>This PR introduces the usage of the new template engine `digAny` custom function. Merging this PR will require the usage of a version of furyctl supports `digAny`.
> See:
> - https://github.com/sighupio/furyctl/pull/640

### Description 📝

Add a new key to the OnPremises kind that allow to disable the installation of the kube-proxy component. By default if the key is not set we keep the current behaviour of enabling the kube-proxy.

Usage:

```yaml
---
kind: OnPremises
metadata:
  name: kube-proxy-less
spec:
  distributionVersion: v1.33.0
  kubernetes:
    ...
    advanced:
      kubeProxy:
        enabled: false
  ...
```

Choosing to disable the kube-proxy will take effect on some parts of the distro, i.e. it will :

- enable kube-proxy-replacement options for Cilium.
- enable eBPF Mode for Calico*.
- remove the kubeproxy monitoring components.

#### \*NOTE 1: Calico support

Calico has some issues of cyclic dependencies with coreDNS when running the data-plane in eBPF mode in the current version of the Tigera Operator included in the networking module v1.38.6. This has been fixed in version 1.40.0 of the Tigera Operator (Calico version: v3.31.0)

See:
- https://github.com/projectcalico/calico/issues/10683

Note that the PR includes 2 commits, reverting the last commit would enable the feature only for Cilium and block disabling the kube-proxy when Calico is the chosen CNI. This could be used if we can't update Calico for some reason and we want to introduce the feature anyway.

#### NOTE 2: Disabling kube-proxy after cluster creation

kube-proxy can be disabled on a live cluster (and re-enabled afterwards). Both Calico and Cilium support being reconfigured to replace the kube-proxy after cluster creation; but I consider this operation too risky to include it at least on a first version of the feature.

One could to the operation manually if desired.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested On-Premises v1.33.1 cluster without any `kubernetes.advaned` configuration set and verified that there are no regressions (kube-proxy and monitoring components are still installed)
- [x] Tested On-Premises v1.33.1 cluster with any `kubernetes.advaned.kubeProxy.enabled: true` configuration set and verified that there are no regressions (kube-proxy and monitoring components are still installed)
- [x] Tested On-Premises v1.33.1 cluster with any `kubernetes.advaned.kubeProxy.enabled: false` configuration set and verified that kube-proxy and monitoring components are not being installed and no errors are produced.
- [x] Tested that a KFDDistribution kind configuration file still produces the right output and does not produce errors.
- [x] Tested that changing kubeProxy.enabled value after cluster has been created is blocked by furyctl.

Tests performed for both Calico and Cilium.

### Future work 🔧

- Setting Cilium in kube-proxy-replacement mode enables to start using other features like l2 announcements, that currently require patching Cilium's config and restarting the cilium-agent daemonset.
  We could handle this by exposing Cilium's config in the schema and handling the restart automatically.
- Handle the migration of enabling and disabling the kube-proxy on a live cluster instead of supporting only creation time. If we really want to support this use case (I don't think so)

### DIFF
--- a/pkg/apis/onpremises/v1alpha2/public/schema.go
+++ b/pkg/apis/onpremises/v1alpha2/public/schema.go
@@ -2,12 +2,10 @@
 
 package public
 
-import (
-	"encoding/json"
-	"fmt"
-	"github.com/sighupio/go-jsonschema/pkg/types"
-	"reflect"
-)
+import "encoding/json"
+import "fmt"
+import "github.com/sighupio/go-jsonschema/pkg/types"
+import "reflect"
 
 type Metadata struct {
 	// The name of the cluster. It will also be used as a prefix for all the other
@@ -131,11 +129,9 @@ type SpecDistributionCustomPatchesConfigMapGeneratorResource struct {
 
 type SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior string
 
-const (
-	SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorCreate  SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "create"
-	SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorMerge   SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "merge"
-	SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorReplace SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "replace"
-)
+const SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorCreate SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "create"
+const SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorMerge SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "merge"
+const SpecDistributionCustomPatchesConfigMapGeneratorResourceBehaviorReplace SpecDistributionCustomPatchesConfigMapGeneratorResourceBehavior = "replace"
 
 type SpecDistributionCustomPatchesConfigMapGeneratorResourceOptions struct {
 	// The annotations of the configmap
@@ -235,11 +231,9 @@ type SpecDistributionCustomPatchesSecretGeneratorResource struct {
 
 type SpecDistributionCustomPatchesSecretGeneratorResourceBehavior string
 
-const (
-	SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorCreate  SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "create"
-	SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorMerge   SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "merge"
-	SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorReplace SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "replace"
-)
+const SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorCreate SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "create"
+const SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorMerge SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "merge"
+const SpecDistributionCustomPatchesSecretGeneratorResourceBehaviorReplace SpecDistributionCustomPatchesSecretGeneratorResourceBehavior = "replace"
 
 type SpecDistributionCustomPatchesSecretGeneratorResourceOptions struct {
 	// The annotations of the secret
@@ -575,11 +569,9 @@ type SpecDistributionModulesAuthProviderBasicAuth struct {
 
 type SpecDistributionModulesAuthProviderType string
 
-const (
-	SpecDistributionModulesAuthProviderTypeBasicAuth SpecDistributionModulesAuthProviderType = "basicAuth"
-	SpecDistributionModulesAuthProviderTypeNone      SpecDistributionModulesAuthProviderType = "none"
-	SpecDistributionModulesAuthProviderTypeSso       SpecDistributionModulesAuthProviderType = "sso"
-)
+const SpecDistributionModulesAuthProviderTypeBasicAuth SpecDistributionModulesAuthProviderType = "basicAuth"
+const SpecDistributionModulesAuthProviderTypeNone SpecDistributionModulesAuthProviderType = "none"
+const SpecDistributionModulesAuthProviderTypeSso SpecDistributionModulesAuthProviderType = "sso"
 
 // Configuration for the Disaster Recovery module.
 type SpecDistributionModulesDr struct {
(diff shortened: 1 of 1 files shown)


templates/distribution/manifests/networking/patches/cilium/kube-proxy-replacement.yaml.tpl
```diff
@@ -0,0 +1,67 @@
...
+  {{- if eq (len $apiAddress) 2 }}
+  {{- $apiPort = $apiAddress._1}}
+  {{- else }}
+      {{- /* If no port is specified in the address we assume the default HTTPS port 443 */ -}}
+      {{- $apiPort = "443" }}
+  {{- end }}
+  {{- if or (eq $apiHost "") (eq $apiPort "") }}
+    {{- fail "Error while calculating the Kubernetes API endpoint host and port. Calculated values are API Host: '$apiHost' and API Port: '$apiPort'" }}
+  {{- end }}
+{{- end }}
```
stefanoghinelli: This compute is here and in the `tigera-kubernetes-service.yaml.tpl` file. Maybe could we extract this logic in the `_helpers.tpl` to avoid maintaining it in two places?

ralgozino: Considering that this logic is only for on-premises and that it should not change much (if ever); I initially did not see the advantage of having it as a helper.

But being that we are going to have a new kind that probably will also support this feature I'm more inclined to move it to a helper.

templates/distribution/manifests/networking/patches/cilium/cilium-config.env.tpl
```diff
@@ -1,12 +1,18 @@
...
-        {{- end }}
+      {{- end }}
     {{- else }}
 cluster-pool-ipv4-cidr={{ .spec.distribution.modules.networking.cilium.podCidr }}
     {{- end }}
-{{- end }}
\ No newline at end of file
+    {{/* Set kube-proxy-replacement only if kube-proxy is disabled */}}
+    {{- if and (hasKeyAny .spec.kubernetes "advanced") (hasKeyAny .spec.kubernetes.advanced "kubeProxy") (not (index .spec.kubernetes.advanced.kubeProxy "enabled")) }}
+kube-proxy-replacement=true
```
speziato: Can the new `IncludeIfKubeProxyEnabled` function be used here? like:
```suggestion
{{ template "IncludeIfKubeProxyEnabled" (dict "state" false "config" . "object" "kube-proxy-replacement=true") }}
```

templates/kubernetes/onpremises/hosts.yaml.tpl
```diff
@@ -156,6 +156,10 @@ all:
         kubernetes_apiserver_certSANs:
 {{ .spec.kubernetes.advanced.apiServerCertSANs | toYaml | indent 10 }}
         {{- end }}
+        {{- if and (hasKeyAny .spec.kubernetes "advanced") (hasKeyAny .spec.kubernetes.advanced "kubeProxy") (not (index .spec.kubernetes.advanced.kubeProxy "enabled")) }}
+        kubeadm_skip_phases:
+          - "addon/kube-proxy"
+        {{- end }}
```
speziato: Same here:
```suggestion
        {{ template "IncludeIfKubeProxyEnabled" (dict "state" false "config" . "object" ("kubeadm_skip_phases: [\"addon/kube-proxy\"]") }}
```

(double check this suggestion, maybe there's a better way to properly create this array)

speziato (changes requested):
Just suggestions to use the new `IncludeIfKubeProxyEnabled` function. Very good functions btw!

templates/distribution/manifests/networking/patches/cilium/cilium-config.env.tpl
ralgozino: Yes it can! but you'll need to remove the {{end}} too.

Anyway, I just realised that the logic for the helper is not 100% correct, it behaves oddly when the kind is not onpremises.

I need to re-check it

speziato: GH does not allow editing the suggestion's lines to include the next line to close the {{ if }} D: If you want I can create a new suggestion, but I won't be offended if my name does not appear in the commit 😂

ralgozino:
I did manage to have a piece of template that would include a piece of text if kubeProxy was in a desired state, but it felt too much like a hack.

This was supposed to be exactly what the `dig` function from Sprig that we include in the template engine should do, but due to how we pass the configuration struct to the template engine it cannot be used (equivalent to the issue with using the `hasKey` function).

Another downside is that the `_helpers.tpl` files with the helpers are phase-scoped, so the helpers for the distribution phase cannot be used in the kubernetes phase templates, for example.

I took the chance and added a new DigAny (following what we did with the `HasKeyAny`) that solves this issue and let us simplify the logic not only for the kubeProxy.enabled detection but potentially simplifying all the rest of the templates where we check for the presence of all the keys.

The downside is that this PR needs that new function to be available in furyctl, otherwise the templates rendering breaks. The CI checks should pass once we have a version of furyctl with the new function.

The PR that introduces the new function is here:
- https://github.com/sighupio/furyctl/pull/640

templates/distribution/manifests/networking/patches/cilium/cilium-config.env.tpl
ralgozino: should be much more cleaner now with the new `digAny` function.

templates/distribution/manifests/networking/kustomization.yaml.tpl
```diff
@@ -17,6 +16,9 @@ resources:
     {{- if eq .spec.distribution.modules.networking.type "calico" }}
   - {{ print $vendorPrefix "/modules/networking/katalog/tigera/on-prem" }}
   - resources/calico-ns.yml
+      {{- if not (.spec | digAny "kubernetes" "advanced" "kubeProxy" "enabled" true) }}
```
speziato: maybe add here the same comment as [cilium-config.env.tpl#R13](https://github.com/sighupio/distribution/pull/459/files#diff-1452617a5c664227166667560563a1f2a1f3f368c217300204953125f2ca44b4R13), just to clarify that we need this condition as it's written
```suggestion
      {{- /* We assume that kubeProxy is enabled by default */}}
      {{- if not (.spec | digAny "kubernetes" "advanced" "kubeProxy" "enabled" true) }}
```

templates/distribution/manifests/networking/kustomization.yaml.tpl
```diff
@@ -50,6 +55,9 @@ patches:
       name: cilium-operator
       namespace: kube-system
     path: patches/cilium/cilium-operator-tolerations.yaml
+    {{- if not (.spec | digAny "kubernetes" "advanced" "kubeProxy" "enabled" true) }}
```
speziato: maybe add here the same comment as [cilium-config.env.tpl#R13](https://github.com/sighupio/distribution/pull/459/files#diff-1452617a5c664227166667560563a1f2a1f3f368c217300204953125f2ca44b4R13), just to clarify that we need this condition as it's written
```suggestion
    {{- /* We assume that kubeProxy is enabled by default */}}
    {{- /* The `digAny` condition needs to be exactly as written below to properly check if the field has been populated */}}
    {{- if not (.spec | digAny "kubernetes" "advanced" "kubeProxy" "enabled" true) }}
```

templates/distribution/manifests/networking/patches/cilium/cilium-config.env.tpl
```diff
@@ -1,12 +1,18 @@
...
+      {{- else }}
 cluster-pool-ipv4-cidr={{ .spec.kubernetes.podCidr }}
-        {{- end }}
+      {{- end }}
     {{- else }}
 cluster-pool-ipv4-cidr={{ .spec.distribution.modules.networking.cilium.podCidr }}
     {{- end }}
-{{- end }}
\ No newline at end of file
+    {{- /* We assume that kubeProxy is enabled by default */}}
```
speziato:
```suggestion
    {{- /* We assume that kubeProxy is enabled by default */}}
    {{- /* The `digAny` condition needs to be exactly as written below to properly check if the field has been populated */}}
```

speziato (dismissed):
LGTM! I left just some minor suggestions to clarify a bit how the `digAny` usage needs to be written to properly check if the field has been populated

ralgozino:
rebased with `main` because of conflict on go models

stefanoghinelli (changes requested):
Hello, I have created the new feat/release-v1.34.0 release branch on the distribution repo and bumped the networking module version to support Tigera v1.40.3.
Here we should: 

- move the base on feat/release-v1.34.0
- resolve any conflicts before merge

stefanoghinelli (approved):
LGTM

ralgozino:
changed target branch. There are no conflicts and unreleased.md was already updated with the new feature information

PR #501, opened 2026-03-20
fix(monitoring): add ScrapeConfig for etcd on dedicated nodes

stefanoghinelli:
### Summary 💡

Fix the etcd Grafana dashboard showing no data when etcd runs on dedicated nodes. When `spec.kubernetes.etcd` is configured, a `ScrapeConfig` resource is now generated to scrape etcd metrics directly from the dedicated etcd nodes.

Closes: https://github.com/sighupio/distribution/issues/500

### Description 📝

See the details of the related issue.

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Deployed an on-premises cluster with external etcd 
- [x] Deployed an on-premises cluster with etcd on control plane nodes for regressions

### Future work 🔧

None.

### DIFF
--- a/templates/distribution/manifests/monitoring/kustomization.yaml.tpl
+++ b/templates/distribution/manifests/monitoring/kustomization.yaml.tpl
@@ -31,9 +31,9 @@ resources:
   - {{ print $vendorPrefix "/modules/monitoring/katalog/haproxy" }}
   - resources/haproxy-scrapeConfig.yaml
     {{- end }}
-    {{- if index .spec.kubernetes "etcd" }}
+  {{- end }}
+  {{- if .spec | digAny "kubernetes" "etcd" nil }}
   - resources/etcd-scrapeConfig.yaml
-    {{- end }}
   {{- end }}
 {{- end }}
 {{- if eq .spec.distribution.common.provider.type "eks" }}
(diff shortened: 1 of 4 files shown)


templates/distribution/manifests/monitoring/kustomization.yaml.tpl
```diff
@@ -67,6 +70,22 @@ resources:
...
+        name: etcd-metrics
+        namespace: kube-system
+  - patch: |-
+      $patch: delete
+      apiVersion: monitoring.coreos.com/v1
+      kind: ServiceMonitor
+      metadata:
+        name: etcd-metrics
+        namespace: kube-system
+{{- end }}
```
ralgozino, line 88: do these patches fail if monitoring type is none? have you tested?

I think the won't because this kustomization.yaml file won't be included but I haven't tested it.

ralgozino (approved):
lgtm, just one doubt to double check but should be ok

templates/distribution/manifests/monitoring/kustomization.yaml.tpl
stefanoghinelli: Yes! The patches work even if the monitoring is "none".

PR #505, opened 2026-04-16
feat(on-prem): support kube-proxy nftables mode

stefanoghinelli:
### Summary 💡

Add support for kube-proxy nftables mode on on-premises clusters via a new field `.spec.kubernetes.advanced.kubeProxy.type`. IPVS mode for kube-proxy is deprecated in Kubernetes v1.35 ([KEP-5495](https://kep.k8s.io/5495)) and will be removed in a future release.

Relates:
- https://github.com/sighupio/installer-on-premises/pull/168

### Description 📝

Introduces `.spec.kubernetes.advanced.kubeProxy.type` (enum: `none` | `ipvs` | `nftables`, default `ipvs`) as the single field. This manages kube-proxy behavior on on-premises clusters and replaces the previous `kubeProxy.enabled` boolean.

### Breaking Changes 💔

- `kubeProxy.enabled` (boolean) is removed and replaced by `kubeProxy.type` (enum). Users with `enabled: false` must set `type: none`; users with `enabled: true` (or no value) must set `type: ipvs` or omit the field (default is `ipvs`).

### Tests performed 🧪

- [x] Verified `nftables-mode.yaml` patch is applied correctly on Calico cluster with `type: nftables`
- [x] Verified `NFTablesMode: Enabled` in Tigera operator logs after patch
- [x] Verified `kube-proxy-replacement` is applied with `type: none` + Cilium
- [x] Verified `kubernetes_kube_proxy_mode` is passed correctly to Ansible for `ipvs` and `nftables`

### Future work 🔧

- furyctl upgrade path (sighupio/furyctl/pull/653)
- Full e2e test with SD v1.35.0-rc.0

---

**EDIT @nutellinoit** — Runtime migration for `kubeProxy.type` (added on top of this branch)

The preflight rule validates the `kubeProxy.type` change but doesn't apply it: on re-apply furyctl regenerates `kubeadm-config`, not the `kube-proxy` ConfigMap, and never restarts the DaemonSet, so the mode change is a silent no-op. I added the missing runtime piece here.

The rule now also blocks `to: ipvs` (deprecated) and carries a `reducers` entry (`lifecycle: post-kubernetes`), so only `ipvs → nftables` and `nil → nftables` pass; `→ none`, `from none`, `→ ipvs` stay blocked. The migration playbook `templates/kubernetes/onpremises/migrations/post-kubernetes.yaml` runs on the control plane: it regenerates the kube-proxy ConfigMap from the kubeadm config (already `mode: nftables`) via `kubeadm init phase addon kube-proxy` and restarts the DaemonSet. Idempotent, no `yq` needed.

On Calico it also moves the Tigera dataplane to nftables in the same phase: kube-proxy on nftables with Calico still on the iptables dataplane can clash, so both must switch together. Note this overlaps with the networking templates, which already set the nftables dataplane when `kubeProxy.type` is nftables (distribution phase). The overlap is intentional and safe: the playbook covers the case where you stop at the kubernetes phase, the templates cover a full create; both set the same value with an idempotent merge, so they converge. Cilium needs no CNI change and is skipped.

On the furyctl side the reducer mechanism is now wired into the onpremises kubernetes phase (runs `migrations/<lifecycle>.yaml` via the ansible runner, driven by the rules; a future migration = new rule + new playbook, no Go), and asks for confirmation before an unsafe migration (`--force migrations` to skip). Requires sighupio/furyctl#679, merge the two together.

Tested e2e on a fresh onprem 1.35 cluster: `ipvs → nftables` switches ConfigMap/runtime to nftables and Calico to `Nftables`; re-apply is a no-op; `→ ipvs`/`→ none` blocked at preflight; the prompt aborts on anything but `yes`.


stefanoghinelli:
Now ready @ralgozino

docs/releases/unreleased.md
```diff
@@ -0,0 +1,20 @@
+# SIGHUP Distribution Release TBD
+
+Welcome to SD release `vTBD`.
+
+The distribution is maintained with ❤️ by the team [SIGHUP by ReeVo](https://sighup.io/).
+
+## New features 🌟
+
+- [[#505](https://github.com/sighupio/distribution/pull/505)] Support NFTables mode for kube-proxy.
```
ralgozino: This description is lacking details, we are not adding support for nftables we are switching the default and this could cause some downtime.

I'm inclined to even move this to breaking changes, because the OS supporting nftables is not granted.

ralgozino (changes requested):
1. Are we tracking somewhere that the upgrade path needs to handle the switch while upgrading? I don't see a related PR.
2. I guess that the underlying OS has to support nftables too. Have we done any research on that?

stefanoghinelli:
Note: the merge must only be done after PR #513

nutellinoit:
**EDIT @nutellinoit** — Runtime migration for `kubeProxy.type` (added on top of this branch)

The preflight rule validates the `kubeProxy.type` change but doesn't apply it: on re-apply furyctl regenerates `kubeadm-config`, not the `kube-proxy` ConfigMap, and never restarts the DaemonSet, so the mode change is a silent no-op. I added the missing runtime piece here.

The rule now also blocks `to: ipvs` (deprecated) and carries a `reducers` entry (`lifecycle: post-kubernetes`), so only `ipvs → nftables` and `nil → nftables` pass; `→ none`, `from none`, `→ ipvs` stay blocked. The migration playbook `templates/kubernetes/onpremises/migrations/post-kubernetes.yaml` runs on the control plane: it regenerates the kube-proxy ConfigMap from the kubeadm config (already `mode: nftables`) via `kubeadm init phase addon kube-proxy` and restarts the DaemonSet. Idempotent, no `yq` needed.

On Calico it also moves the Tigera dataplane to nftables in the same phase: kube-proxy on nftables with Calico still on the iptables dataplane can clash, so both must switch together. Note this overlaps with the networking templates, which already set the nftables dataplane when `kubeProxy.type` is nftables (distribution phase). The overlap is intentional and safe: the playbook covers the case where you stop at the kubernetes phase, the templates cover a full create; both set the same value with an idempotent merge, so they converge. Cilium needs no CNI change and is skipped.

On the furyctl side the reducer mechanism is now wired into the onpremises kubernetes phase (runs `migrations/<lifecycle>.yaml` via the ansible runner, driven by the rules; a future migration = new rule + new playbook, no Go), and asks for confirmation before an unsafe migration (`--force migrations` to skip). Requires sighupio/furyctl#679, merge the two together.

Tested e2e on a fresh onprem 1.35 cluster: `ipvs → nftables` switches ConfigMap/runtime to nftables and Calico to `Nftables`; re-apply is a no-op; `→ ipvs`/`→ none` blocked at preflight; the prompt aborts on anything but `yes`.

nutellinoit (approved):
LGTM

PR #510, opened 2026-05-04
feat: add kapp group for nginx

stefanoghinelli:
### Summary 💡

Sequence the creation of the Forecastle `Ingress` after the Ingress NGINX controller's rolling update, using kapp annotations. This avoids the `validate.nginx.ingress.kubernetes.io` admission failure (`host "..." and path "/" is already defined`) that can happen when an Ingress with the same host has just been deleted via kapp but the first `delete` request isn't propagated in time.

It also updates the e2e tests for the KFDDistribution provider.

Closes:

Relates:

### Description 📝

See summary.

### Breaking Changes 💔

None. 

### Tests performed 🧪

- [x] EKS/KFDDistribution e2e test passed
- [x] OnPrem manual testing

### Future work 🔧

None


davide-rossi-ce (approved):
LGTM!

PR #511, opened 2026-05-05
fix(onpremises): allow arbitrary keys under auth.overrides.ingresses

smerlos:
### Summary 💡

The OnPremises Go templates already consume `.spec.distribution.modules.auth.overrides.ingresses.pomerium.host` through the generic auth ingress helpers and the Pomerium-specific templates. However, `schemas/public/onpremises-kfd-v1alpha2.json` declared `auth.overrides.ingresses` as a closed object whose only allowed keys were `dex` and `gangplank`, with `additionalProperties: false`.

As a result, schema validation rejected any OnPremises configuration that overrode the Pomerium ingress, even though the rendered manifests already support that override.

This PR aligns the OnPremises schema with the convention already used by `ekscluster` and `kfddistribution`: `auth.overrides.ingresses` becomes an open map with `additionalProperties` referencing the existing `Spec.Distribution.Modules.Auth.Overrides.Ingress` schema.

It also adds the missing OnPremises schema fixtures and bats runner so this path is covered by schema tests, and wires the new runner into the Drone `qa` pipeline.

Closes:

Relates:

### Description 📝

**Why the change is correct.** `ekscluster` and `kfddistribution` already declare `Spec.Distribution.Modules.Auth.Overrides.properties.ingresses` as an open map:

```json
{
  "type": "object",
  "additionalProperties": {
    "$ref": "#/$defs/Spec.Distribution.Modules.Auth.Overrides.Ingress"
  }
}
```

OnPremises was the only distribution declaring the same field as a closed object with `additionalProperties: false` and explicit `dex` / `gangplank` properties. The generated Go types reflected the same divergence: `ekscluster` and `kfddistribution` generated `SpecDistributionModulesAuthOverridesIngresses` as `map[string]SpecDistributionModulesAuthOverridesIngress`, while OnPremises generated a closed struct with only `Dex` and `Gangplank`.

This PR makes OnPremises match its sibling distributions.

**Why the change is needed.** The templates already handle auth ingress overrides generically by package name. Pomerium specifically uses `.spec.distribution.modules.auth.overrides.ingresses.pomerium.host` to render the Pomerium Ingress host (`pomerium-ingress.yml.tpl`), Pomerium environment configuration (`pomerium-config.env.tpl` → `FORWARD_AUTH_HOST`, `AUTHENTICATE_SERVICE_HOST`), and Dex OIDC redirect URI (`dex.yml.tpl`). Before this change, users could not pass a schema-valid OnPremises config that used the Pomerium override already implemented by the templates.

**What is generated, what is hand-written.** The schema change, fixtures, bats runner, and `.drone.yml` step are hand-written. `pkg/apis/onpremises/v1alpha2/public/schema.go` and `docs/schemas/onpremises-kfd-v1alpha2.md` are generated with:

```
mise run _generate-go-models
mise run format-go
mise run generate-docs
```

**Note on additional codegen drift.** Running the documented codegen pipeline also re-flowed imports in three pre-existing generated files (`pkg/apis/ekscluster/v1alpha2/private/schema.go`, `pkg/apis/ekscluster/v1alpha2/public/schema.go`, `pkg/apis/kfddistribution/v1alpha2/public/schema.go`). These are import-ordering fixes from `gci` that the previous tree had drifted on; the changes are isolated to the codegen commit so the next person to run the pipeline gets a clean tree. Codegen is idempotent on the resulting tree.

### Breaking Changes 💔

**JSON Schema** — none. Configurations that previously validated with `dex` and/or `gangplank` ingress overrides still validate. This change only allows additional auth ingress keys that conform to the same ingress override schema.

**Go API** — minor generated type shape change. The exported generated type `pkg/apis/onpremises/v1alpha2/public.SpecDistributionModulesAuthOverridesIngresses` changes from a closed struct to `map[string]SpecDistributionModulesAuthOverridesIngress`. External Go consumers importing this generated type directly must migrate from field access such as `.Dex` / `.Gangplank` to map indexing such as `["dex"]`, `["gangplank"]`, or `["pomerium"]`.

This matches the generated type shape already used by `ekscluster/v1alpha2/public` and `kfddistribution/v1alpha2/public`.

### Tests performed 🧪

- [x] `jq empty schemas/public/onpremises-kfd-v1alpha2.json`
- [x] `git diff -- schemas/public/onpremises-kfd-v1alpha2.json` shows only the `auth.overrides.ingresses` block changing from closed `properties` to open `additionalProperties`.
- [x] Direct `jv` validation of `tests/schemas/public/onpremises-kfd-v1alpha2/001-ok.yaml` exits 0.
- [x] Direct `jv` validation of `tests/schemas/public/onpremises-kfd-v1alpha2/002-ok.yaml` exits 0.
- [x] Direct `jv` validation of `tests/schemas/public/onpremises-kfd-v1alpha2/003-no.yaml` exits non-zero with the expected path `/spec/distribution/modules/auth/overrides/ingresses/pomerium` and message containing `missing property` and `'ingressClass'`.
- [x] `bats -t tests/e2e/onpremises/schema.sh` (3/3 ok)
- [x] `bats -t tests/e2e/kfddistribution/schema.sh` (24/24 ok — no regression)
- [x] `mise run license-check` clean
- [x] Replicated Drone `schema-check` step: `cat schemas/public/ekscluster-kfd-v1alpha2.json | json-patch -p schemas/private/ekscluster-kfd-v1alpha2.patch.json | jq -r > /tmp/eks-private.json && diff schemas/private/ekscluster-kfd-v1alpha2.json /tmp/eks-private.json` → no diff.
- [x] Replicated Drone `test-create-config`: `furyctl create config --kind OnPremises --version v1.34.1 --name test --distro-location $(pwd) --disable-analytics --no-tty --workdir /tmp/test-onpremises` succeeded; the generated `furyctl.yaml` validates against the updated schema with `jv`.
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `mise run lint-go` (Issues before processing: 6112, after processing: 0)
- [x] Codegen idempotence: re-ran `mise run _generate-go-models && mise run format-go && mise run generate-docs`; `git status` clean.

### Future work 🔧

- Move shared schema-test helpers from the distro-specific bats runners (`tests/e2e/kfddistribution/schema.sh`, `tests/e2e/onpremises/schema.sh`) into a common `tests/e2e/helpers.bash` to reduce duplication.
- Backfill more OnPremises schema fixtures to approach the broader schema coverage currently present for EKS fixtures (001–013). This PR adds the directory, runner, and the focused fixtures needed to cover the Pomerium ingress override bug.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -75,6 +75,7 @@ steps: # qa
     commands:
       - eval "$(mise activate bash --shims)"
       - bats -t tests/e2e/kfddistribution/schema.sh
+      - bats -t tests/e2e/onpremises/schema.sh
 
   - name: test-create-config
     image: quay.io/sighup/mise:v2025.4.4
--- a/docs/schemas/onpremises-kfd-v1alpha2.md
+++ b/docs/schemas/onpremises-kfd-v1alpha2.md
@@ -782,58 +782,9 @@ Override the common configuration with a particular configuration for the Auth m
 
 ## .spec.distribution.modules.auth.overrides.ingresses
 
-### Properties
-
-| Property                                                             | Type     | Required |
-|:---------------------------------------------------------------------|:---------|:---------|
-| [dex](#specdistributionmodulesauthoverridesingressesdex)             | `object` | Optional |
-| [gangplank](#specdistributionmodulesauthoverridesingressesgangplank) | `object` | Optional |
-
 ### Description
 
 Override the definition of the Auth module ingresses.
-
-## .spec.distribution.modules.auth.overrides.ingresses.dex
-
-### Properties
-
-| Property                                                                      | Type     | Required |
-|:------------------------------------------------------------------------------|:---------|:---------|
-| [host](#specdistributionmodulesauthoverridesingressesdexhost)                 | `string` | Required |
-| [ingressClass](#specdistributionmodulesauthoverridesingressesdexingressclass) | `string` | Required |
-
-## .spec.distribution.modules.auth.overrides.ingresses.dex.host
-
-### Description
-
-Use this host for the ingress instead of the default one.
-
-## .spec.distribution.modules.auth.overrides.ingresses.dex.ingressClass
-
-### Description
-
-Use this ingress class for the ingress instead of the default one.
-
-## .spec.distribution.modules.auth.overrides.ingresses.gangplank
-
-### Properties
-
-| Property                                                                            | Type     | Required |
-|:------------------------------------------------------------------------------------|:---------|:---------|
-| [host](#specdistributionmodulesauthoverridesingressesgangplankhost)                 | `string` | Required |
-| [ingressClass](#specdistributionmodulesauthoverridesingressesgangplankingressclass) | `string` | Required |
-
-## .spec.distribution.modules.auth.overrides.ingresses.gangplank.host
-
-### Description
-
-Use this host for the ingress instead of the default one.
-
-## .spec.distribution.modules.auth.overrides.ingresses.gangplank.ingressClass
-
-### Description
-
-Use this ingress class for the ingress instead of the default one.
 
 ## .spec.distribution.modules.auth.overrides.nodeSelector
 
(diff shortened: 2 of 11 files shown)


smerlos:
The suggested approach is to replicate the current on-premise schema (adding the Pomerium item) to the EKSInstaller and KFDDistribution provider instead of just leaving it open.

PR #514, opened 2026-05-22
feat(immutable/schema)!: add directories, remove multipath and iscsi

smerlos:
## Summary

Reworks the `Immutable` v1alpha2 node schema to align it with what is actually rendered by the Butane templates and consumed by `furyctl`:

- **Adds** declarative directory creation via `spec.infrastructure.nodes[].storage.directories[]`, mirroring the Butane Flatcar v1.1.0 `storage.directories` object (`path` required; `overwrite`, `mode`, `user{id,name}`, `group{id,name}` optional).
- **Removes** `spec.infrastructure.nodes[].multipath` — not rendered by any template, not referenced by `furyctl`, and natively handled by Flatcar via sysext + systemd units (already expressible through `spec.infrastructure.nodes[].systemd`).
- **Removes** `spec.infrastructure.nodes[].iscsi` — not rendered by any template, not referenced by `furyctl`. The iSCSI initiator is configured via chained ignition in the first-stage PXE boot, outside the fury schema.

## Why

`multipath` and `iscsi` were schema-only fields with no rendering path — they created false expectations that `furyctl` would apply them. Removing them shrinks the API surface to what we actually deliver. `storage.directories` plugs a real gap: today users need `systemd-tmpfiles` or ad-hoc scripts to create paths that downstream modules assume exist (e.g., `/var/lib/etcd`, `/var/lib/containerd-data`, `/etc/cni/net.d`).

### Schema (`schemas/public/immutable-kfd-v1alpha2.json`)
- **ADD** `$defs.Spec.Infrastructure.Node.Storage.Directory` (object with `path` required + `overwrite`, `mode`, `user`, `group` optional).
- **ADD** `$defs.Spec.Infrastructure.Node.Storage.properties.directories` (array of `Directory`).
- **REMOVE** `$defs.Spec.Infrastructure.Node.Multipath` and the corresponding `multipath` property on `Node`.
- **REMOVE** `$defs.Spec.Infrastructure.Node.Iscsi` and the corresponding `iscsi` property on `Node`.

### Butane templates (`templates/infrastructure/immutable/butane/`)
Adds the standard rendering block in `controlplane.bu.tpl`, `etcd.bu.tpl`, `worker.bu.tpl`, `loadbalancer.bu.tpl`:

```gotemplate
  # User-provided additional directories
{{- if hasKeyAny .node.storage "directories" }}
  directories:
{{ .node.storage.directories | toYaml | indent 4 }}
{{- end }}
```

### Go models (`pkg/apis/**/schema.go`)
Regenerated via `mise run generate-go-models`. Net effect on `pkg/apis/immutable/v1alpha2/public/schema.go`: new `SpecInfrastructureNodeStorageDirectory` struct; `SpecInfrastructureNodeMultipath` and `SpecInfrastructureNodeIscsi` removed.

**Side effect**: the same `generate-go-models` task also regenerates `pkg/apis/ekscluster/v1alpha2/{public,private}/schema.go`, `pkg/apis/kfddistribution/v1alpha2/public/schema.go` and `pkg/apis/onpremises/v1alpha2/public/schema.go`. These four files were stale in `feat/kind-immutable` (committed Go did not match the corresponding JSON schemas), so the regeneration adds 1, 38, 13 and 39 new types respectively — all derived from existing JSON definitions, **0 types removed**. The cleanup is reconciliation, not a schema change.

### Docs and fixtures
- `docs/schemas/immutable-kfd-v1alpha2.md` regenerated via `mise run generate-docs`.
- `tests/schemas/public/immutable-kfd-v1alpha2/002-ok.yaml` and `examples/immutable-full-example.yaml` cleaned of `multipath:` and `iscsi:` blocks (the example/test scaffolds previously defined those fields per node).

## Breaking changes

Any `furyctl.yaml` (kind `Immutable`) that defines `spec.infrastructure.nodes[].multipath` or `spec.infrastructure.nodes[].iscsi` will fail validation with `additional property not allowed`. Go importers of `SpecInfrastructureNodeMultipath` or `SpecInfrastructureNodeIscsi` will fail at compile time.

### Migration

- **multipath**: declare a `multipathd.service` unit under `spec.infrastructure.nodes[].systemd.units` and ship `/etc/multipath.conf` via `storage.files`. Reference: https://www.flatcar.org/docs/latest/setup/storage/multipath/
- **iscsi**: use chained ignition in the PXE first stage to configure `/etc/iscsi/initiatorname.iscsi`, `iscsid.conf` and the `iscsid.service` unit before Flatcar installs to disk. Alternatively, use `spec.infrastructure.nodes[].systemd.units` + `storage.files`.

## Verification

- [x] `jq . schemas/public/immutable-kfd-v1alpha2.json > /dev/null`
- [x] `mise run generate-go-models`
- [x] `mise run generate-docs`
- [x] `mise run lint-go`
- [x] `mise run format-go`
- [x] `mise run license-check`
- [x] `go test ./...` — `pkg/apis/config` ok, rest no test files (autogen)
- [x] Downstream consumer `furyctl` (via `replace` directive against this branch): `go build ./...`, `go vet ./...`, `go test ./...` all pass (14/14 packages, 0 FAIL)

## Notes

- `furyctl` requires **no changes** — it does not reference any of the affected fields. Verified with `grep -rn -E "iscsi|multipath|StorageDirectory" furyctl/{internal,pkg,cmd}` (0 matches) plus full build/vet/test on `furyctl` with the local `replace` directive pointing to this branch.
- No `rules/immutable-kfd-v1alpha2.yaml` entries added for `storage.directories` — same convention as the existing `storage.links` field, which has none.

### DIFF
--- a/docs/schemas/immutable-kfd-v1alpha2.md
+++ b/docs/schemas/immutable-kfd-v1alpha2.md
@@ -84,6 +84,7 @@ The name of the cluster. It will also be used as a prefix for all the other reso
 | [infrastructure](#specinfrastructure)           | `object` | Required |
 | [kubernetes](#speckubernetes)                   | `object` | Required |
 | [plugins](#specplugins)                         | `object` | Optional |
+| [toolsConfiguration](#spectoolsconfiguration)   | `object` | Optional |
 
 ## .spec.distribution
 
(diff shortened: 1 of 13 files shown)


ralgozino (approved):
lgtm!

PS: not a breaking change. We haven't released anything yet ;-)

PR #524, opened 2026-06-12
feat(immutable): pre and post Install commands

ralgozino:
### Summary 💡

Support specifying `preInstall` commands that will be run before installing flatcar to disk and `postInstall` commands that will be run after Flatcar has been installed to disk and before rebooting the machine.

The install-flatcar step now also reports to furyctl when it can't install Flatcar to disk because a disk with the ROOT label was already present.

Relates:
- https://github.com/sighupio/furyctl/pull/668

### Description 📝

See summary

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested flatcar installation with pre and post install commands
- [x] Tested flatcar installation with pre install commands
- [x] Tested flatcar installation with post install commands
- [x] Tested flatcar installation without pre or post install commands

### Future work 🔧

None

### DIFF
--- a/docs/schemas/immutable-kfd-v1alpha2.md
+++ b/docs/schemas/immutable-kfd-v1alpha2.md
@@ -5434,11 +5434,13 @@ Defines the bare metal infrastructure configuration, including nodes, network bo
 
 ### Properties
 
-| Property                                                | Type      | Required |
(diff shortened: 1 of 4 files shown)


smerlos (approved):
LGTM

Issue #522, opened 2026-06-10
[on-prem] Support dnsZone per host override in hosts.yaml template


stefanoghinelli:
Is your feature request related to a problem? Please describe

When deploying a cluster with both on-premises and cloud nodes, all hosts share the same dnsZone value defined in spec.kubernetes.dnsZone.
This makes it impossible to assign different FQDNs to hosts that belong to different DNS zones, for example controlplane.k8s.local for on-prem nodes and etcd01.k8s.cloud for cloud nodes.

Describe the solution you'd like

Add support for a per-host dnsZone override in the furyctl.yaml, for masters, etcd, workers and loadBalancers host entries. When set, the local value takes precedence over the global spec.kubernetes.dnsZone when computing kubernetes_hostname and etcd_initial_cluster.

Example:
```
spec:
  kubernetes:
    dnsZone: k8s.local
    masters:
      hosts:
        - name: controlplane-0
          ip: 192.168.2.2
    etcd:
      hosts:
        - name: etcd-0
          ip: 192.168.2.6
          dnsZone: k8s.cloud
```

Additional context

The change would require:
1. Updating the hosts.yaml.tpl template to use default $dnsZone (index $h "dnsZone") (or equivalent) at each point where the hostname FQDN is composed.
2. Updating on-premises schema to allow the optional dnsZone field under host objects.

PR #530, opened 2026-06-24
feat(on-prem): support dnsZone per host override

stefanoghinelli:
### Summary 💡

Adds support for a per-host `dnsZone` override in `furyctl.yaml` for `masters`, `etcd`, `nodes`, and `loadBalancers` host entries. 

Closes: [#522](https://github.com/sighupio/distribution/issues/522)

### Description 📝

When deploying a cluster with both on-premises and cloud nodes, all hosts previously shared the single `spec.kubernetes.dnsZone` value. This made it impossible to assign different FQDNs to hosts belonging to different DNS zones (example: `controlplane.k8s.local` for on-prem nodes and `etcd01.k8s.cloud` for cloud nodes).

The new optional `dnsZone` field can be set on any host entry:

```yaml
spec:
  kubernetes:
    dnsZone: k8s.local
    masters:
      hosts:
        - name: master-0
          ip: 192.168.1.1
        - name: master-1
          ip: 192.168.1.2
          dnsZone: k8s.cloud   # overrides the global dnsZone for this host only
    etcd:
      hosts:
        - name: etcd-0
          ip: 192.168.1.5
          dnsZone: k8s.cloud
```

### Breaking Changes 💔

None. 

### Tests performed 🧪

-  [x] Tested a clean provisioning with a local distro-location pointing to this branch.

### Future work 🔧

None.

### DIFF
--- a/tests/schemas/public/onpremises-kfd-v1alpha2/005-ok.yaml
+++ b/tests/schemas/public/onpremises-kfd-v1alpha2/005-ok.yaml
@@ -1,0 +1,77 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+# Tests the following cases:
+
+# Given a valid OnPremises config
+# And some host entries under masters, etcd, workers, and loadBalancers have a per-host "dnsZone" field
+# When I validate the config against the schema
+# Then no errors are returned
+
+---
+apiVersion: kfd.sighup.io/v1alpha2
+kind: OnPremises
+metadata:
+  name: onpremises-test-per-host-dnszone
+spec:
+  distributionVersion: v1.35.0
+  kubernetes:
+    pkiFolder: /etc/pki
+    ssh:
+      username: user
+      keyPath: ~/.ssh/id_rsa
+    dnsZone: k8s.local
+    controlPlaneAddress: api.k8s.local
+    podCidr: 10.0.0.0/16
+    svcCidr: 10.1.0.0/16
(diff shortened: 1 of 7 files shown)


tests/schemas/public/onpremises-kfd-v1alpha2/005-ok.yaml
ralgozino, line 1: this test has exactly the same zone in all nodes, it would not catch if you are taking the zone of a different node by mistake.

It would be better to have different zones on each node and test that the right zone is picked (and maybe also leave at least one node without the value set so you can check it picks the global value?)

ralgozino (changes requested):
changes look good, a note on the tests

nutellinoit (approved):
LGTM! Ship it 🚀

tests/schemas/public/onpremises-kfd-v1alpha2/005-ok.yaml
stefanoghinelli: Do you intend to use separate zone values for each host and keep the nodes without overwriting them to handle the global fallback, or are you asking for a functional test that runs the hosts.yaml and verifies that each host receives the correct zone in the output?

ralgozino: I assumed that you were checking that the right values was being passed on, but on a second closer look I see that you are just validating the schema (that is possible to set the per-host dnsZone values). The test is ok as-is.

For future reference, is better to use the one of the IANA reserved domains for documentation, test, etc.: https://www.iana.org/help/example-domains

ralgozino (approved):

Issue #523, opened 2026-06-10
[on-prem] kubernetes_image_registry set only in master.vars, worker nodes use wrong sandbox_image registry


stefanoghinelli:
Describe the bug

When spec.kubernetes.advanced.registry is set in furyctl.yaml, the generated hosts.yaml places kubernetes_image_registry only under master.vars. Worker nodes never receive this variable, so the containerd role falls back to the hardcoded default registry.sighup.io/fury/on-premises when rendering sandbox_image for the pause container.

To Reproduce

1. Configure a cluster with a custom registry in furyctl.yaml:
```
spec:
  kubernetes:
    advanced:
      registry: "custom-registry.example.com/fury/on-premises"
```
2. Run furyctl apply
3. Inspect the generated .furyctl/<cluster>/kubernetes/hosts.yaml, kubernetes_image_registry appears only under master.vars
4. After provisioning, SSH into a worker node and check containerd config /etc/containerd/config.toml
5. The sandbox_image points to registry.sighup.io/fury/on-premises/pause:x.y instead of custom registry

Expected behavior

kubernetes_image_registry should be available to all nodes in the cluster. The containerd role uses it to compute sandbox_image (pause container) for every node group. It should be placed under all.vars in the generated hosts.yaml.

Screenshots

N/A

Software versions

- SIGHUP Distribution version: v1.32.2
- furyctl version:  v0.34.1

Additional context

The fix is to move the kubernetes_image_registry rendering from master.vars to all.vars in the template. This is particularly impactful in air-gap envs.

PR #532, opened 2026-06-24
fix(on-premises): move kubernetes_image_registry to all.vars

stefanoghinelli:
### Summary 💡

When spec.kubernetes.advanced.registry is set in furyctl.yaml, the generated hosts.yaml was placing kubernetes_image_registry under master.vars only. Worker nodes never received the variable, causing the containerd role to fall back to the hardcoded registry.sighup.io/fury/on-premises when computing sandbox_image for the pause container.

Closes: https://github.com/sighupio/distribution/issues/523

###  Description 📝

See the issue.

###  Breaking Changes 💔

None.

###  Tests performed 🧪

- [x] Generated hosts.yaml and verified kubernetes_image_registry appears under all.vars

###  Future work 🔧

None.

PR #534, opened 2026-06-24
fix(on-prem): propagate kubernetes_image_registry to worker and etcd nodes

nutellinoit:
### Summary 💡

On-prem clusters with a custom registry (`spec.kubernetes.advanced.registry`): worker/infra nodes were pulling the `pause` image from the default registry instead.

Closes: #523

### Description 📝

`kubernetes_image_registry` was set only under the `master` group vars in the generated `hosts.yaml`, so worker/infra and external etcd nodes fell back to the hardcoded `registry.sighup.io/fury/on-premises` for `sandbox_image`. Added it to the `nodes` and external-`etcd` group vars too. Left out of `haproxy` on purpose (L4 LB, not a cluster node).

### Breaking Changes 💔

None. With `advanced.registry` unset the output is unchanged.

### Tests performed 🧪

- [x] Template structure checked (mirrors the existing `master` block).

### Future work 🔧

None.

### DIFF
--- a/templates/kubernetes/onpremises/hosts.yaml.tpl
+++ b/templates/kubernetes/onpremises/hosts.yaml.tpl
@@ -176,6 +176,9 @@ all:
       vars:
         dns_zone: "{{ $dnsZone }}"
         etcd_initial_cluster: "{{ $etcdInitialCluster | join "," }}"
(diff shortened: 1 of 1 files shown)


ralgozino:
isn't this the same as https://github.com/sighupio/distribution/pull/532 ?

Issue #315, opened 2024-11-22
`alertmanager.installDefaultRules` name may be confusing


ralgozino:
In https://github.com/sighupio/fury-distribution/pull/159 we introduced the field `.spec.distribution.modules.monitoring.alertmanager.installDefaultRules`.

The `installDefaultRules` name of this field can be confusing to the user because, even when the field is false, Prometheus Rules are installed in the cluster, triggering alerts.

What this field disables is KFD default Alertmanager configuration for sending **notifications** for the aforementioned alerts.

Setting this field to true, let's the user create a custom Alertmanager configuration for routing the alerts.

I suggest renaming the field to something like `installDefaultConfig` instead.

PR #537, opened 2026-06-25
fix(templates/monitoring): AlertmanagerConfigs conditional deploy

### LINKED ISSUE #536: Alert PrometheusOperatorRejectedResources triggered with default config
### Describe the bug

In a cluster with monitoring enabled, if the user does not set the URLs for the Slack and DeadmanSwitch webhooks, the `PrometheusOperatorRejectedResources ` alert will be always triggered because the secrets with the URL are missing.

### To Reproduce
1. Create a cluster with monitoring enabled and `alertmanager.installDefaultRules` enabled (default value).
3. Wait some minutes a see that in Alertmanager the alert `PrometheusOperatorRejectedResources` is triggered saying that:

```
Prometheus operator in monitoring namespace rejected 3 alertmanager/AlertmanagerConfig resources
```

4. Set the webhooks URLs in the furyctl.yaml and see the alert go away.

### Expected behavior

No alerts should be triggered if I haven't configured the Webhooks URLs.

### Screenshots

N/A

### Software versions

 - SIGHUP Distribution version: v1.34.0
 - furyctl version: 0.34.1

### Additional context

N/A

ralgozino:
### Summary 💡

Deploy the AlertmanagerConfigs included with the monitoring module only if needed.

Fixes #536

Related:
- https://github.com/sighupio/distribution/issues/315

### Description 📝

Deploy the AlertmanagerConfigs included with the monitoring module for Slack and DeadmanSwith webhooks only when the URLs for those services are set.

If the URLs are _not_ set in the configuration file, the secrets with the URL are not created, resulting in a configuration error on Prometheus Operator that triggers a `PrometheusOperatorRejectedResources` alert.

Added logic to deploy the AlertmanagerConfigs CRs only when the URLs are set, to avoid false postivies.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] setting the webhook URL - no alert
- [x] unsetting the webhook URL: alertmanagerConfig is not deployed
- [x] setting `installDefaultRules: false` -> alertmanagerConfigs are not deployed even if webhooks URLs are set, `furyctl apply` works without errors
- [x] setting `installDefaultRules: true` explicitly and the webhook URLs works

### Future work 🔧

With this change the `installDefaultRules` field does not make much sense any more. We could deprecate it, also its name and description do not match well with what it actually does.

I suggest deprecating the field, if the URLs are not set (default state) now the AlertmanagerConfig CR will not be created, having the same effect as setting `installDefaultRules` to `false`.

### DIFF
--- a/docs/releases/v1.35.0.md
+++ b/docs/releases/v1.35.0.md
@@ -25,6 +25,7 @@ This version adds support for Kubernetes 1.35 and updates modules.
 
 - [[[#515](https://github.com/sighupio/distribution/issues/515)]] **Don't try to delete Kyverno Policies when CRDs are not present**: the distribution pre-apply script tried to delete the default Kyverno policies even when the CRDs were not present in the cluster. Now the script checks for the CRDs before attempting to delete them.
 - [[#532](https://github.com/sighupio/distribution/pull/532)] **Fix worker nodes not receiving custom registry in OnPremises clusters**: `kubernetes_image_registry` was placed under `master.vars` instead of `all.vars` in the generated `hosts.yaml`, causing worker nodes to fall back to the default `registry.sighup.io/fury/on-premises` registry for the pause container (`sandbox_image`).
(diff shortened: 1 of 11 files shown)


stefanoghinelli (approved):
lgtm

PR #539, opened 2026-06-25
feat(kfd): pin ansible and its build toolchain for the mise-managed install

nutellinoit:
### Summary 💡

Adds `ansible` under the `onpremises` and `immutable` provider tool sections so `furyctl` installs a self-contained, versioned Ansible (+ galaxy collections) via the bundled mise, instead of requiring it on the host. Each section also pins the build toolchain (`python` + `uv`) mise uses to build the venv.

Closes:

Relates: https://github.com/sighupio/furyctl/pull/681

### Description 📝

Ansible is only used by the `OnPremises` and `Immutable` kinds, so it is pinned per provider (mirroring how `tools.eks` groups the EKS-only tools) rather than under `tools.common`. Each provider section pins ansible-core, its build toolchain and the galaxy collections — the block is repeated under both `onpremises` and `immutable`:

```yaml
tools:
  onpremises:
    ansible:
      version: 2.21.0
      python: "3.12"
      uv: "0.9"
      collections:
        - { name: ansible.posix, version: 2.2.0 }
        - { name: community.general, version: 13.0.1 }
  immutable:
    ansible: # same block as onpremises
```

`python` and `uv` live inside `ansible` on purpose: the right versions depend on the ansible version (ansible-core 2.21 needs python ≥ 3.11), so the distribution that picks ansible must pick its toolchain too. Both are optional — furyctl falls back to its defaults when a distro pins only `version` + `collections`. Full mechanism in the furyctl PR.

### Breaking Changes 💔

None — the sections are optional; distros without them keep using the host Ansible.

### Tests performed 🧪

- [x] On-prem e2e on Vagrant (furyctl run inside a Linux container): `apply` creates a v1.35.5 cluster, playbooks run by the mise-installed Ansible 2.21.0 + collections from `tools.onpremises`
- [x] Self-contained install verified on a clean Linux container (no host python/uv/pipx)
- [x] EKSCluster/KFDDistribution don't pull ansible; distros without the sections fall back to host Ansible (verified against a real v1.34.1)

### Future work 🔧

- Pin the same block on other distribution lines as they adopt the mise-managed Ansible.

- [x] I've considered all the different kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change
- [x] CI is green


ralgozino (approved):
lgtm

PR #544, opened 2026-06-30
feat(immutable): add upgrade capability to immutable kind

smerlos:
### Summary 💡

Add the immutable **install** templates for the v0.35.0 release: version-vars `group_vars`, per-node arch in
the inventories + butane, and the upgrade playbooks. Couple
`kfd.immutable.version` to the SD minor (1.35.5), matching `onpremises`.

Closes:

Relates: sighupio/furyctl#688, sighupio/installer-immutable#8, installer-immutable tag `v1.35.5-rc.0`

### Description 📝

- `group_vars/all.yml.tpl` — render role version vars from furyctl's `.versions` context (single source).
- `hosts.yaml.tpl` (infra + kubernetes) — per-node `node_arch` from the config.
- butane `update-server-config` (`_helpers.tpl` + the 4 `.bu.tpl`) — Flatcar OS-update config on every node.
- `upgrade-{control-plane,etcd,worker-nodes}` playbooks. (Cert renewal uses the release's existing `renew-certificates.yaml`.)
- `kfd.yaml` — `immutable.version` 1.34.8 → **1.35.5** to couple with the SD minor (= `onpremises`).

All additive; no existing release content changed.

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] `format-go` clean, license headers present; templates render via furyctl `validate config`
- [x] End-to-end install + upgrade 1.34.8 → 1.35.5 (control-plane first, then per-node workers) → 6/6,
      validated on an onprem-consistent setup
- [ ] Fresh-install recreate against this exact branch — pending

### Future work 🔧

Immutable upgrade-path dirs named in SD/distributionVersion strings (like `configs/upgrades/onpremises/`),
driven by the common distributionVersion diff — added when a prior SD release ships immutable 1.34.x.

- [ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)
- [x] I've considered all the different cluster kinds — this change is immutable-scoped and leaves OnPremises / EKSCluster / KFDDistribution untouched
- [ ] CI is green


templates/kubernetes/immutable/upgrade-etcd.yaml.tpl
```diff
@@ -0,0 +1,38 @@
...
+    # furyctl copies admin.conf into this work dir only in the later core phase, so a clean upgrade run (no
+    # leftover from a prior apply, e.g. after rm -rf .furyctl or without --skip-deps-download) has no
+    # kubeconfig for the localhost-delegated kubectl checks. Fetch it from a control plane up front; 54, 55
+    # and 56 share this work dir, so the health gate, drain and uncordon all find ./admin.conf.
+    - name: Fetch admin.conf to the controller for the upgrade run
+      run_once: true
+      ansible.builtin.fetch:
+        src: /etc/kubernetes/admin.conf
+        dest: ./admin.conf
+        flat: true
```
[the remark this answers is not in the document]
smerlos: Fixed in 00c942a9: the inline fetch now delegates to a reachable control-plane via the upgrade-gates role (fetch_admin_conf.yml), so it works on dedicated etcd and tolerates an unreachable control-plane instead of failing on the play host. Thanks for the catch.

templates/kubernetes/immutable/upgrade-etcd.yaml.tpl
```diff
@@ -0,0 +1,38 @@
...
+  hosts: {{ if $etcdOnControlPlane }}control_plane{{ else }}etcd{{ end }}
+  serial: 1
+  become: true
+  vars:
+    etcd_upgrade: true
+  pre_tasks:
+    # furyctl copies admin.conf into this work dir only in the later core phase, so a clean upgrade run (no
+    # leftover from a prior apply, e.g. after rm -rf .furyctl or without --skip-deps-download) has no
+    # kubeconfig for the localhost-delegated kubectl checks. Fetch it from a control plane up front; 54, 55
+    # and 56 share this work dir, so the health gate, drain and uncordon all find ./admin.conf.
```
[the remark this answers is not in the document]
smerlos: Fixed in 00c942a9: removed the non-existent 54/55/56 numbering. The comment was replaced with a note that admin.conf lives only on control-planes and is fetched via the upgrade-gates role — it doesn't reference the three upgrade plays by name.

smerlos:
@ralgozino — companion to sighupio/installer-immutable#8: `00c942a9` moves the admin.conf fetch into the upgrade-gates role and delegates it to a control-plane, fixing the dedicated/external-etcd case you reported. You can continue with the review.

templates/infrastructure/immutable/ansible/group_vars/all.yml.tpl
ralgozino, line 1: I'm wondering, what is the benefit of having these variables here instead of in the hosts.yaml with the rest of the variables?

Also, the haproxy image and tag can be set by the user in the furyctl.yaml file, the value is then passed to the hosts.yaml file under the load_balancers group vars but this is overriding it.

templates/kubernetes/immutable/group_vars/all.yml.tpl
ralgozino, line 1: similar to the same file for the infrastructure phase. This file should not be needed anymore IMO and we can just set the vars in the hosts.yaml file.

Notice that there are some duplicated vars between this file and the hosts.yaml (kubernetes_version, for example) and some vars that are not needed in the kubernets phase (haproxy_*, for example)

templates/kubernetes/immutable/upgrade-control-plane.yml.tpl
```diff
@@ -0,0 +1,63 @@
...
+# license that can be found in the LICENSE file.
+---
+# Control-plane upgrade, serial 1: refresh the kubernetes sysext, renew certs, and kubeadm upgrade apply (restarts the static pods in-band); the post_tasks reboot activates the staged OS (and the swapped kubelet binary, unless its config-patch handler restarted the kubelet first).
+- name: Upgrade the control plane
+  hosts: control_plane
+  serial: 1
+  become: true
+  vars:
+    upgrade: true
+    k8s_upgrade: true
```
ralgozino: leaving a comment here as reminder: isn't the variable to trigger upgrades in the `kube-*` roles called `update`? What is this variable piloting?

ralgozino (changes requested):

templates/kubernetes/immutable/upgrade-control-plane.yml.tpl
smerlos: There is no update var. k8s_upgrade was redundant with upgrade, so I removed it. The plays now only set upgrade: true.

templates/infrastructure/immutable/ansible/group_vars/all.yml.tpl
smerlos: Moved inline to hosts.yaml; the user override now wins and the immutable.yaml pin is the fallback.

templates/kubernetes/immutable/group_vars/all.yml.tpl
smerlos: Moved inline to hosts.yaml; group_vars emptied and the duplicate kubernetes_version removed.

kfd.yaml
```diff
@@ -22,7 +22,7 @@ kubernetes:
     installer: v1.35.5-rc.1
   immutable:
     version: 1.35.5
-    installer: v1.35.5-rc.0
+    installer: v1.35.5-rc.16
```
ralgozino: FYI: you can point to a branch here instead of a tag while developing. So you don't have to tag everytime you want to test something

templates/kubernetes/immutable/upgrade-control-plane.yml.tpl
ralgozino: Yes, I meant `upgrade` sorry

templates/infrastructure/immutable/ansible/group_vars/all.yml.tpl
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+---
+# Intentionally empty. The immutable.yaml version vars now render inline in hosts.yaml so a user
+# override in furyctl.yaml wins (a group_vars/all.yml file outranks inline inventory vars and would
+# silently shadow it). Rendered empty rather than deleted so a stale file from a previous furyctl
+# version is overwritten on re-apply.
```
ralgozino: No need to have backwards compatiblity for this file, this was never merged

templates/kubernetes/immutable/group_vars/all.yml.tpl
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+---
+# Intentionally empty. The immutable.yaml version vars now render inline in hosts.yaml so a user
+# override in furyctl.yaml wins (a group_vars/all.yml file outranks inline inventory vars and would
+# silently shadow it). Rendered empty rather than deleted so a stale file from a previous furyctl
+# version is overwritten on re-apply.
```
ralgozino: this was never merged, there are no previous furyctl versions to take care of

templates/kubernetes/immutable/hosts.yaml.tpl
```diff
@@ -194,8 +200,25 @@ all:
     kubernetes_kubeconfig_path: ./
     kubernetes_version: "{{ .kubernetes.version }}"
     kubernetes_local_pki_dir: "{{ .spec.kubernetes.pkiPath }}/master"
-    {{- if (.spec.kubernetes | digAny "advanced" "registry" false) }}
-    kubernetes_image_registry: "{{ .spec.kubernetes.advanced.registry }}"
+    # kubernetes_image_registry: user value from furyctl.yaml wins; immutable.yaml pin is the fallback.
+    kubernetes_image_registry: "{{ .spec.kubernetes | digAny "advanced" "registry" "" | default .versions.kubernetes_image_registry }}"
+    # Version pins from immutable.yaml (single source); consumed by containerd / kube-control-plane / os-upgrade / sysext roles.
+    containerd_sandbox_image: {{ .versions.containerd_sandbox_image }}
```
ralgozino: if the user sets kubernetes_image_registry to use a custom registry. the containerd_sandbox_image won't be using that custom registry (and it should) because is read from the immutable.yaml file

templates/kubernetes/immutable/upgrade-control-plane.yml.tpl
```diff
@@ -0,0 +1,76 @@
...
+  hosts: control_plane
+  gather_facts: false
+  become: false
+  tasks:
+    - name: Remove the fetched super-admin.conf from the controller
+      run_once: true
+      delegate_to: localhost
+      ansible.builtin.file:
+        path: ./super-admin.conf
+        state: absent
```
ralgozino: why delete the super-admin.conf file from the control-plane node? this does no look right to me

ralgozino (changes requested):

templates/kubernetes/immutable/upgrade-control-plane.yml.tpl
smerlos: This only removes the fetched local copy on the controller or bastion, not the CP node file. I renamed it to make that clear and made deletion optional with upgrade_delete_fetched_super_admin, default false. By default we keep it for infra ops, with a TODO to revisit the security risk.

ralgozino: We agreed to delete this task and we will harden the .furyctl folder afterwards

templates/kubernetes/immutable/apply.yaml.tpl
```diff
@@ -20,7 +20,7 @@
     - etcd
 {{- end }}
 
-# TODO: que pasa cuando no quiero usar keepalived entonces no lo especifico en el furyctl.yaml?
+# TODO: make keepalived optional — today the VIP role always runs even if the user omits keepalived from furyctl.yaml.
```
ralgozino: This TODO comment is old, if the VIP is disabled the tasks are properly skipped

smerlos: Comment Removed.

kfd.yaml
```diff
@@ -19,10 +19,10 @@ kubernetes:
     installer: v3.5.0-rc.0
   onpremises:
     version: 1.35.5
-    installer: v1.35.5-rc.2
+    installer: v1.35.5-rc.1
   immutable:
-    version: 1.34.8
-    installer: v1.34.8-rc.0
+    version: 1.35.5
+    installer: v1.35.5-rc.16
```
ralgozino:
```suggestion
    installer: v1.35.5
```

ralgozino (changes requested):
every thing is looking good 🏁 ❤️ 

well done!

Let's wait until we release the final version of immutable installer to merge this PR.

ralgozino (approved):
let's go 🎉

PR #547, opened 2026-07-01
fix(immutable): set node booted after sshd

ralgozino:
### Summary 💡

Make the report node as booted systemd unit depend on SSHD.

### Description 📝

Report node as booted after sshd has started, so furyctl (via Ansible) can reach the nodes.

If we do it after the network is online we could fall in the case where the SSH daemon is not started yet and furyctl tried to connect to it, failing the apply procedure.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested successful Immutable kind cluster creation

### Future work 🔧

None

~~- [] I've updated the `docs/releases/unreleased.md` file (or equivalent)~~ N.A.
- [x] CI is green

### DIFF
--- a/templates/infrastructure/immutable/butane/_helpers.tpl
+++ b/templates/infrastructure/immutable/butane/_helpers.tpl
@@ -271,6 +271,7 @@ passwd:
         Description=Report node status to furyctl
         Requires=network-online.target
         After=network-online.target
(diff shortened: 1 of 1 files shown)


smerlos (approved):
LGTM

PR #562, opened 2026-07-22
Fix/immutable kubelet configuration

ralgozino:
### Summary 💡

Make kubelet configuration schema for Immutable kind accept additional properties, aligning it with OnPremises.

### Description 📝

Make kubelet configuration schema for Immutable kind accept additional properties, aligning it with OnPremises.

The field had only two hardcoded properties of the 89 possible values.

Kept maxPods as documented field and added the systemReserved field and subfields to the documented properties because they are the most used ones and an editor can auto-complete them (and are validated by the json schema)

### Breaking Changes 💔

The field now accepts more values than before, but this is the expected behaviour.

### Tests performed 🧪

- [x] Tested setting the kubeletParameters field for an Immutable cluster and validated that they get applied.
- [x] Tested not setting kubeletParameters field for an Immutable cluster and validated that the apply works.

### Future work 🔧

Decide if we want to actually validate all 89 possible values for the kubeletConfiguration field (that depend on kubeadm and kubernetes versions, probably)

> [!IMPORTANT]
>

- [x] CI is green

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -10,6 +10,7 @@ Welcome to the latest release of SD maintained by SIGHUP by ReeVo team.
 ## Bug fixes 🐞
 
 - [[#555](https://github.com/sighupio/distribution/pull/555)] Monitoring templates: don't try to patch the alertmanagerConfigs when they are not being deployed at all.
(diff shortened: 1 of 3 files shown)


smerlos (approved):
LGTM

stefanoghinelli (approved):
lgtm

PR #566, opened 2026-07-23
feat(immutable/butane): derive initramfs kernel args for no-DHCP nodes

### LINKED ISSUE #565: Immutable: a node on a segment without DHCP cannot finish booting
### Describe the bug

Immutable nodes boot over the network. Very early in boot (in the initramfs) a node downloads its sysext images (containerd, kubelet, ...) over HTTP from the boot server. The network configuration that furyctl writes only starts working later, after switch-root.

When the segment has a DHCP server, the initramfs gets an IP address on its own, so the download works. When there is no DHCP server (the manual OS-ISO install case), the node has no address that early, the download never succeeds, and the node hangs forever retrying.

### To Reproduce

1. On a segment with no DHCP server, install an Immutable node from the OS ISO using a static `network` configuration.
2. Boot the node from disk.
3. See the node stay in the initramfs retrying the sysext download; it never reports `booted`.

### Expected behavior

A node with a static network configuration should boot without needing a DHCP server.

### Software versions

- SIGHUP Distribution version: v1.35.0
- furyctl version: 0.35.0

### Additional context

The network settings only take effect after switch-root, so they do not help the initramfs. The node needs its network configured from the kernel command line (an `ip=` / `nameserver=` kernel argument) to have connectivity early enough to download the sysext images.

smerlos:
### Summary 💡

Nodes on a segment with no DHCP server could not finish booting. This makes furyctl fill in the boot-time network setting for such a node automatically, and adds an optional per-node `kernelArguments` field so you can set kernel arguments yourself.

Closes: #565

Relates:

### Description 📝

For the problem this solves, see the linked issue.

What this PR does:

- **Auto-derive (default).** If a node has a static IP and no DHCP, furyctl now adds one boot-time kernel argument (`ip=...`, plus `nameserver=...`) so the node has a working network early enough to download its sysext images. It picks the interface that has a gateway and adds **exactly one** `ip=` — two would make dracut give up and power the machine off, leaving it unrecoverable. Nodes that use DHCP get nothing extra, exactly like before.
- **Override.** A new optional node field `kernelArguments` lets you set kernel arguments yourself. It has two optional lists, `shouldExist` and `shouldNotExist`, exactly like Butane's `kernel_arguments`. When set, furyctl uses your values instead of deriving them.

Both lists are optional, matching Butane's `kernel_arguments` (rendered as `should_exist` / `should_not_exist`).

### Breaking Changes 💔

None. `kernelArguments` is a new optional field, and nodes that use DHCP render exactly as before.

### Tests performed 🧪

- [x] Schema validation, covering `kernelArguments` with only `shouldExist`, with both lists, with a wrong-type value, and with an unknown key — the valid ones pass and the invalid ones are rejected.
- [x] `furyctl create config --kind Immutable --version v1.35.0` still works.
- [x] Rendered a real 8-node config and checked the resulting Ignition kernel arguments for each scenario:
  - override with both lists → both rendered
  - override with only `shouldExist` → only `should_exist` rendered
  - override with only `shouldNotExist` → only `should_not_exist` rendered
  - override with empty lists → no `kernel_arguments` block
  - auto, one static interface with a gateway → one `ip=` + `nameserver=`
  - auto, DHCP interface → nothing added
  - auto, two static interfaces → exactly one `ip=` (the one with the gateway)
  - auto, gateway on the second interface → that interface is the one chosen

### Future work 🔧

- Reject `dhcp4: true` together with `addresses` on the same interface in the schema (today the static address is silently dropped).
- Warn when a static interface has no `nameservers` and the boot server is referenced by hostname.
- Document real-hardware interface naming (the `eth0` name only matches on virtio NICs).
- Update the `installer-immutable` manual-ISO docs to describe the now-working no-DHCP boot.

- [x] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change — this change is scoped to Immutable only
- [ ] CI is green

### DIFF
--- a/templates/infrastructure/immutable/butane/_helpers.tpl
+++ b/templates/infrastructure/immutable/butane/_helpers.tpl
@@ -99,6 +99,58 @@ passwd:
 {{- end }}
 {{- end }}
 
+{{- /* initramfs network kargs so Ignition can fetch sysext before switch-root; without them a
+node on a DHCP-less segment never boots. node.kernelArguments, if set, is emitted verbatim
+(manual override); otherwise ip=/nameserver= are derived from the node's static interfaces.
+Emit exactly ONE ip= (the gateway iface): dracut can't split two and powers off, and first_boot
+re-adds removed args, bricking the node. networkd configures the rest after switch-root. Node
+configs only, never install-flatcar (ignition-kargs-helper hard-fails on PXE). Read map keys
+with index, not $ka.foo: templates run missingkey=error, so an absent key aborts the render. */}}
+
+{{- define "initramfs-kargs" }}
+{{- if index .node "kernelArguments" }}
+{{- $ka := index .node "kernelArguments" }}
+{{- $se := index $ka "shouldExist" }}
+{{- $sne := index $ka "shouldNotExist" }}
+{{- if or $se $sne }}
+kernel_arguments:
+{{- if $se }}
+  should_exist:
+    {{- range $se }}
+    - {{ . }}
+    {{- end }}
+{{- end }}
+{{- if $sne }}
+  should_not_exist:
+    {{- range $sne }}
+    - {{ . }}
+    {{- end }}
+{{- end }}
+{{- end }}
+{{- else }}
+{{- $selName := "" }}
+{{- $sel := dict }}
+{{- range $name, $iface := (.node.network | digAny "ethernets" dict) }}
+{{- if and (not (index $iface "dhcp4")) (index $iface "addresses") }}
+{{- if or (eq $selName "") (and (not (index $sel "gateway")) (index $iface "gateway")) }}
+{{- $selName = $name }}
+{{- $sel = $iface }}
+{{- end }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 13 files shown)


templates/infrastructure/immutable/butane/_helpers.tpl
```diff
@@ -99,6 +99,58 @@ passwd:
...
+
+kernel_arguments:
+  should_exist:
+    - ip={{ first $cidr }}::{{ $sel | digAny "gateway" "" }}:{{ last $cidr }}:{{ $.node.hostname }}:{{ $selName }}:none
+{{- range ($sel | digAny "nameservers" "addresses" list) }}
+    - nameserver={{ . }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
```
ralgozino, line 152: This can probably be refactored and simplified, but being that we need to add the kernel argument for downloading the official sysexts from our servers, we can simplify it then

ralgozino (approved):
🚀

PR #568, opened 2026-07-24
fix(immutable): set search domain

ralgozino:
### Summary 💡

Update butane templates to include the search domain if defined in the node's network configuration. This value was being ignored.

### Description 📝

When setting the search domain in an interface with static IP configuration, the search domain was not being actually set in the node's networkd configuration.

This PR updates the butane template helper to include the search domain if it is defined.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested that if the search domain is set, the configuration ends up in the node and it actually works (e.g. pinging the host without the fqdn resolves the fqdn appending the search domain)
- [x] Tested that there are no errors and no search domain is set when the field is not set in the furyctl.yaml file.

### Future work 🔧

None

> [!IMPORTANT]
>

- [X] I've updated the `docs/releases/unreleased.md` file (or equivalent)
- [x] CI is green


smerlos (approved):
LGTM, Nice catch! 🤟

PR #569, opened 2026-07-24
feat(immutable)!: add node storage.disks and storage.filesystems

### LINKED ISSUE #556: Immutable: storage.additionalDisks passes validation but is never rendered (all node roles)

smerlos:
### Summary 💡

Immutable nodes can now set up extra disks through `spec.infrastructure.nodes[].storage.disks` and `storage.filesystems`,
mirroring Butane's own sections like `storage.files` already does. They replace `storage.additionalDisks`, which
validated but was never rendered.

Closes: #556

Relates:

### Description 📝

`storage.additionalDisks` was dead: it passed validation, no template read it, so extra disks were silently never
partitioned.

It could not be wired up as it was. Its shape was not Butane's: it nested the filesystem inside each partition, while
Butane keeps disks and filesystems as two sibling lists joined by a device path. So the node config now speaks Butane
directly. Both keys are optional and independent, so you can partition without formatting, or format a whole disk with no
`disks` entry.

Subfields are declared inline with `additionalProperties: false`, like `storage.files`. They come from the Butane version
furyctl uses, checked in both directions: nothing Butane accepts is missing, nothing we accept is unknown to it.

This is not only style. Butane merely warns on an unknown key and furyctl keeps going, so a `sizeMib` typo used to pass
validation, get dropped, and give you a partition sized "as large as possible" on a node that boots fine. It now fails
early.

The four node templates get the same block. Dropping `additionalDisks` also removes the four `$defs` that only it used.

Known limit: the link between a partition label and a filesystem device is a plain path, and neither Butane nor JSON
Schema checks it. A typo there still only shows up at boot.

### Breaking Changes 💔

`storage.additionalDisks` is removed. A `furyctl.yaml` that still sets it now fails validation. No running cluster is
affected, since the field was never rendered.

Migration is mechanical:

| before | after |
|---|---|
| `additionalDisks[].device` | `disks[].device` |
| `partitions[].sizeMiB` | `disks[].partitions[].size_mib` |
| `partitions[].filesystem` | its own `filesystems[]` entry, with `device: /dev/disk/by-partlabel/<label>` |
| `filesystem.mountPoint` | `filesystems[].path` |
| `filesystem.mountOptions` | `filesystems[].mount_options` |
| the mount it implied | `with_mount_unit: true` |

### Tests performed 🧪

- [x] Schema suite green (45/45). Five new cases cover the typos and shapes that used to slip through: singular `disk`,
      the old nested `filesystem`, `sizeMib`, a format outside the enum, and a disk with no device. Removing the new
      subfields makes four of them pass again, so they test what they claim.
- [x] Fields checked one by one against Butane, both ways. Required fields were probed with `butane`, not guessed.
- [x] The block the templates emit goes through `butane --strict` into valid Ignition, no warnings, mount unit created.
      `disks` alone and `filesystems` alone both work.
- [x] A real pre-migration config is rejected with a per-node error and migrates without loss.
- [x] CI is green.

Not covered: a live `furyctl apply`. The block is identical in the four templates and was validated through `butane`
directly.

### Future work 🔧

- `tests/templates/immutable/` does not exist, so the regression suite never renders these templates.
- `systemd.units` declares no subfields, so it validates nothing — the same gap this PR closes for disks, left out to
  keep the scope clean.
- `Types.DevicePath` rejects hyphens, so it cannot express `/dev/disk/by-id/...`. That is why `device` is a plain string.

- [x] CI is green

### DIFF
--- a/schemas/public/immutable-kfd-v1alpha2.json
+++ b/schemas/public/immutable-kfd-v1alpha2.json
@@ -356,18 +356,148 @@
     "Spec.Infrastructure.Node.Storage": {
       "type": "object",
       "additionalProperties": false,
-      "description": "Storage configuration for the node, including install disk and additional disks with partitions.",
+      "description": "Storage configuration for the node, including the install disk and the Butane storage sections.",
       "properties": {
         "installDisk": {
           "$ref": "#/$defs/Types.DevicePath",
           "description": "The disk device where the OS will be installed. Example: /dev/sda"
         },
-        "additionalDisks": {
+        "disks": {
           "type": "array",
+          "description": "The list of disks to be partitioned. Every disk must have a unique device. See https://www.flatcar.org/docs/latest/provisioning/config-transpiler/configuration/",
           "items": {
-            "$ref": "#/$defs/Spec.Infrastructure.Node.Storage.Disk"
-          },
-          "description": "Additional disks to partition and mount."
+            "type": "object",
+            "additionalProperties": false,
+            "description": "Represents a disk to be partitioned.",
+            "properties": {
+              "device": {
+                "type": "string",
+                "description": "The absolute path to the device. Devices are typically referenced by the /dev/disk/by-* symlinks. Example: /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_drive1"
+              },
+              "wipe_table": {
+                "type": "boolean",
+                "description": "Whether or not the partition tables shall be wiped. When true, the partition tables are erased before any further manipulation. Otherwise, the existing entries are left intact. Defaults to false."
+              },
+              "partitions": {
+                "type": "array",
+                "description": "The list of partitions to be created on the disk. Every partition must have either a number or a label.",
+                "items": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "description": "Represents a partition to be created on the disk.",
+                  "properties": {
+                    "label": {
+                      "type": "string",
+                      "description": "The PARTLABEL for the partition, used to reference it from a filesystem as /dev/disk/by-partlabel/LABEL. Cannot contain a colon. Example: data",
+                      "maxLength": 36,
+                      "pattern": "^[^:]*$"
+                    },
+                    "number": {
+                      "type": "integer",
+                      "description": "The partition number, which will change the partition table entry at that location. If not specified, the next available partition number will be used."
+                    },
+                    "size_mib": {
+                      "type": "integer",
+                      "description": "The size of the partition with a unit of mebibytes. If zero, the partition will be made as large as possible."
+                    },
+                    "start_mib": {
+                      "type": "integer",
+                      "description": "The start of the partition with a unit of mebibytes. If zero, the partition will be positioned at the start of the largest block available."
+                    },
+                    "type_guid": {
+                      "type": "string",
+                      "description": "The GPT partition type GUID. If omitted, the default will be 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (Linux filesystem data).",
+                      "pattern": "^(|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})$"
+                    },
+                    "guid": {
+                      "type": "string",
+                      "description": "The GPT unique partition GUID.",
+                      "pattern": "^(|[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})$"
+                    },
+                    "wipe_partition_entry": {
+                      "type": "boolean",
+                      "description": "If true, Ignition will clobber an existing partition if it does not match the config. If false, Ignition will fail instead. Defaults to false."
+                    },
+                    "should_exist": {
+                      "type": "boolean",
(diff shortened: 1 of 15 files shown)


schemas/public/immutable-kfd-v1alpha2.json
ralgozino, line 1: why drop the fields schema? Is fine taking the values as they come from the furyctl.yaml configuration file, but if we already have the schema for the subfields let's keep them (as we did for the other sections)

ralgozino (changes requested):
PR description does not follow our template, checklist items missing check

schemas/public/immutable-kfd-v1alpha2.json
smerlos: I had assumed `additionalDisk` was a Butane key — it isn't (`warning: unused key additionalDisk`), so it was a wrapper of ours that matched nothing downstream.

Dropped it: `storage.disks` and `storage.filesystems` are now siblings of `storage.files`, with Butane's own field names and their subfields declared inline.

ralgozino (changes requested):
the changes strictly related to the new `disks` property look good, but the PR is showing me a lot of unrelated changes too - like the changes of my PR #568 that was merged on main.

May be you did a merge (that went wrong) instead of a rebase? 🤔

smerlos:
> the changes strictly related to the new `disks` property look good, but the PR is showing me a lot of unrelated changes too - like the changes of my PR #568 that was merged on main.
> 
> May be you did a merge (that went wrong) instead of a rebase? 🤔

I rebased again and updated the pr title and description.

ralgozino (approved):
lgtm

FYI
> Future work 🔧
> `tests/templates/immutable/` does not exist, so the regression suite never renders these templates.

this is no longer true since #558 that added the baseline for immutable

PR #570, opened 2026-07-24
fix(immutable): remove old fields

ralgozino:
### Summary 💡

Clean up the Immutable public schema by removing fields and type definitions that were introduced in early iterations of the schema but never implemented in the distribution templates or consumed by `furyctl`.

Relates:
- #568 

### Description 📝

This PR updates the Immutable schema by removing unused fields and renaming some others.

**Removed (unused, confirmed not referenced by templates or furyctl):**
- **Bond networking**: the `bonds` property on `Spec.Infrastructure.Node.Network` and the `Spec.Infrastructure.Node.Network.Bond` definition. The butane renderer (`templates/infrastructure/immutable/butane/_helpers.tpl`) only iterates `network.ethernets`. Users can use the `files` directive to write the needed configuration files for bonding (and other advanced network configurations).
- **Orphan `Types.*` definitions** that were never `$ref`'d anywhere: `Types.Duration`, `Types.EnvRef`, `Types.FileRef`, `Types.SemVer`, `Types.SshPubKey`, `Types.TcpPort`.

**Other tidy-ups:**
- Renamed `Spec.Infrastructure.KernelParameter` → `Spec.Infrastructure.Node.KernelParameter` (node-level parameter), having a global variable does not make sense because it will affect the Load Balancers, Control Plane and Worker machines equally. Users can use YAML anchors to achieve the same result and avoid repetition.
- Dropped the "bonds, and VLANs" mention from the `Node.Network` description.
- Updated schema tests by removing bond references.

**Explicitly kept:**
- `members.ip` — retained (currently ignored by templates, reserved for future use).

### Breaking Changes 💔

Removed global `kernelParameters` and `nodes.network.bonds` parameters from the schema (weren't working before, so actual change).

### Tests performed 🧪

Example:

- [x] Tested a successful Immutable cluster creation from scratch

### Future work 🔧

- Implement or formally remove the reserved `members.ip` field.

> [!IMPORTANT]
>

- [X] My branch is up-to-date with the target branch and there are no conflicts
- [X] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change
- [x] CI is green


smerlos (changes requested):
Tested this: schema is clean after the removal (108 `$defs`, nothing dangling), 39/39 bats pass, and a cluster built from scratch with this branch came up 6/6 Ready with no `bond` in the render. `bonds:` is now properly rejected. Nice cleanup.

One thing, and it's only the changelog.

**The global `kernelParameters` is still there.** Breaking Changes says it was removed, but it's at `immutable-kfd-v1alpha2.json:111` — the diff only changes its `$ref`:

```diff
         "kernelParameters": {
           "items": {
-            "$ref": "#/$defs/Spec.Infrastructure.KernelParameter"
+            "$ref": "#/$defs/Spec.Infrastructure.Node.KernelParameter"
```

`validate config` still accepts it and the value never reaches the render. A node-level `kernelParameters` does land as `sysctl_parameters`, so it's just this field that isn't wired.

It was already dead before this PR, so nothing broke — the note just announces a removal that didn't happen. The rename suggests you meant to drop it (the global now points at a `$def` called `...**Node**.KernelParameter`).


Also hit some pre-existing issues while testing, unrelated to this — filed as #573.

ralgozino:
Thanks, I was sure I had removed the global field but it seems I got confused. I've actually removed it and updated the tests.

I also took the chance to update the schema docs.

smerlos (approved):
❤️

PR #576, opened 2026-07-28
feat: release v1.35.1

ralgozino:
### Summary 💡

Prepare for releasing SIGHUP Distribution v1.35.1.

### Description 📝

- Update `kfd.yaml` with new `1.35.1` distro version and bump installer immutable to `v1.35.5-rev.2`
- Update documentation to target 1.35.1 (readme, release notes, etc.)
- Update CI and regression and e2e tests to use 1.35.1 as target version for installation and upgrades
- bump `mise` in CI to a newer version

### Breaking Changes 💔

None introduced by this PR.

### Tests performed 🧪

- [x] Tested cluster creation and upgrade for Immutable from v1.35.0 -> v1.35.1

### Future work 🔧

`e2e-onpremises-upgrades` pipeline is skipped every run. It's the only pipeline that `depends_on` a trigger-gated pipeline (`e2e-onpremises`) instead of `qa`, which Drone skips even when the parent succeeds.

> [!IMPORTANT]
>

- [x] CI is green kinda:
  - [x] QA pipeline is green
  - [x] kfddistribution e2e are green
  - [x] onpremises creation e2e are green
  - [ ] onpremises _upgrade_ e2e are being skipped for some reason (was like this from before)
  - [x] eks alinux e2e are green
  - [ ] the other 2 eks e2e test failed with too many hosted zones error after applying the distribution phase


nutellinoit (changes requested):
Nice work, just a few leftovers before we ship. `tests/e2e/ekscluster-upgrades/manifests/furyctl-upgrade-version-1.35.1.yaml` was only renamed, inside it's still `distributionVersion: v1.35.0` (it works due to --distro-location but it's wrong)

Same thing in `tests/e2e/kfddistribution/manifests/furyctl-8-migrate-from-sso-to-none.yaml`, still on 1.35.0 while all its 13 siblings are on 1.35.1, so the migration chain downgrades halfway through.

ralgozino:
> Nice work, just a few leftovers before we ship. `tests/e2e/ekscluster-upgrades/manifests/furyctl-upgrade-version-1.35.1.yaml` was only renamed, inside it's still `distributionVersion: v1.35.0` (it works due to --distro-location but it's wrong)

Fixed

> Same thing in `tests/e2e/kfddistribution/manifests/furyctl-8-migrate-from-sso-to-none.yaml`, still on 1.35.0 while all its 13 siblings are on 1.35.1, so the migration chain downgrades halfway through.

Fixed. Note that there was now downgrade because the `--upgrade` flag is not set when applying that file and the `--distro-location` is still set and **the test is commented out** so is not run:
https://github.com/sighupio/distribution/blob/d220850dc23bbea50064b8612fdec22ac3487bd4/tests/e2e/kfddistribution/e2e-kfddistribution.sh#L76

smerlos (dismissed):
❤️

PR #577, opened 2026-07-29
fix(templates): fetch admin.conf from whichever master actually has it

stefanoghinelli:
### Summary 💡

Makes the admin.conf fetch in `verify-playbook.yaml` pick the right master regardless of Ansible's host ordering, and stops the playbook from failing on purpose to signal "cluster doesn't exist", so furyctl can tell that case apart from a real failure.

Relates:
- sighupio/furyctl#717
- #387

### Description 📝

`verify-playbook.yaml` used a `fail:` task as its only way to tell furyctl "no master has admin.conf yet". Furyctl only ever checked whether the playbook succeeded or failed, so any other kind of failure (a host unreachable, a checksum mismatch during the fetch) looked identical to "cluster doesn't exist" on the furyctl side.

This PR:
- Removes the `fail:` task. The playbook now completes cleanly whenever nothing goes wrong, whether or not a master has admin.conf.
- Picks the master to fetch from explicitly (`admin_conf_host`, using the same aggregation #387 introduced), and fetches with `delegate_to` instead of relying on `run_once`'s default host selection.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested a cluster creation. Still correctly recognized as "cluster doesn't exist".
- [x] Tested a plain `apply` on an existing cluster.
- [x] Tested a real `--upgrade` with a `distributionVersion` change. Diff still correctly detected, real kubeadm upgrade path runs.
- [x] Tested a forced local failure (fetch destination not reachable). Now fails with an explicit error instead of "cluster doesn't exist".
- [x] Tested a forced Ansible failure. Fails with an explicit error.
- [x] Tested adding a 4th control-plane node in `masters.hosts`, simulating an immutable edit. Cluster still correctly recognized as existing and the immutability check on `masters.hosts.*` correctly blocked the change.

### Future work 🔧

Apply the same fix to the Immutable kind

> [!IMPORTANT]
>

- [x] CI is green

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -16,6 +16,8 @@ Welcome to the latest release of SD maintained by SIGHUP by ReeVo team.
 - [[#561](https://github.com/sighupio/distribution/pull/561)] Immutable: align possible properties for Kubernetes phase `kubeletConfiguration` parameter with OnPremises, accepting all possible values now.
 - [[#564](https://github.com/sighupio/distribution/pull/564)] OnPremises: fixed a race in the preflight `verify-playbook.yaml` where fetching `admin.conf` from multiple masters in parallel could randomly fail with a checksum mismatch.
 - [[#568](https://github.com/sighupio/distribution/pull/568)] Immutable: actually set the search domains for an interface if it is defined in the node configuration, this value was being ignored until now.
+- [[#577](https://github.com/sighupio/distribution/pull/577)] Makes the admin.conf fetch in verify-playbook.yaml pick the right master, and stops the playbook from failing on purpose to signal "cluster doesn't exist".
+
 
 ## Breaking Changes 💔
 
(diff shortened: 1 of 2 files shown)


ralgozino (commented):
the proposed changes lgtm. But, the verify-playbook is not verifying anything any more isn't it? Should we rename it?

stefanoghinelli:
Renaming it here would require a backward-compatible change or a different approach in furyctl; this would expand the scope of this fix. I suggest to keep the filename as is for now and address the renaming issue separately.

ralgozino (approved):
Ok, but beware that this is intentional technical debt. Let's track it somewhere.

I don't think it is too much work to fix, check if the new file exists otherwise fallback to the old name, another approach could be to use the distribution version to determine the playbook name.

stefanoghinelli:
Follow up sighupio/distribution#595 sighupio/furyctl#757

PR #579, opened 2026-07-30
fix(immutable): default route

### LINKED ISSUE #578: immutable: default route ignored in node configuration
### Describe the bug
Specifying a default network route using the `to: default` keyworkd does not create the actual default route.

### To Reproduce
1. In an immutable cluster configuration file, define a default route for a node using the `default` keyword as the route destination (proposed by the schema), for example:
```yaml
spec:
  infrastructure:
    nodes:
      - hostname: host.example.com
        network:
          ethernets:
            eth00:
              # gateway: 192.168.0.1
              routes:
                - to: default
                  via: 192.168.0.1
```

2. Apply the configuration
4. Log in into the machine with ssh and list the routes with `ip route show`, `routel`,  or similar.
5. See that the default route is missing.

### Expected behavior
The default route should be set.

### Screenshots
NA

### Software versions

 - SIGHUP Distribution version: v1.35.0
 - furyctl version: 0.35.0

### Additional context
networkd seems to not accept `default` as a destination and silently drops the route. Using `0.0.0.0/0` instead of `default` works.

ralgozino:
### Summary 💡

Convert `default` into `0.0.0.0/0` for networkd compatibility.

Closes #578 

### Description 📝

We are exposing interfaces configuration in the schema following netplan's format because it has a nicer UX and let us inspect easily the node's network configuration from furyctl and the templates.

We then generate networkd configuration files from the netplan format. The templates were passing to networkd the `default` keyword for static routes but networkd silently ignores it instead of creating the default route. networkd accepts `0.0.0.0/0` instead of `default`.

This PR converts `default` into `0.0.0.0/0` for networkd compatibility.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Declared a default route and checked on the node that the default route is present
- [x] Declared several routes and checked that they are all present in the node

### Future work 🔧

None

> [!IMPORTANT]
>

- [x] CI is green

### DIFF
--- a/docs/releases/v1.35.1.md
+++ b/docs/releases/v1.35.1.md
@@ -25,6 +25,7 @@ No changes in modules version.
 - [[#561](https://github.com/sighupio/distribution/pull/561)] Immutable: align possible properties for Kubernetes phase `kubeletConfiguration` parameter with OnPremises, accepting all possible values now.
 - [[#564](https://github.com/sighupio/distribution/pull/564)] OnPremises: fixed a race in the preflight `verify-playbook.yaml` where fetching `admin.conf` from multiple masters in parallel could randomly fail with a checksum mismatch.
 - [[#568](https://github.com/sighupio/distribution/pull/568)] Immutable: actually set the search domains for an interface if it is defined in the node configuration, this value was being ignored until now.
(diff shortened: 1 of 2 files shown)


smerlos (approved):
🤟

PR #580, opened 2026-07-31
fix(eks/templates): prevent HAProxy LoadBalancer Service recreation on EKS

stefanoghinelli:
### Summary 💡

Prevent HAProxy `LoadBalancer` Services on EKS from being deleted and recreated, with a new NLB, on some `furyctl apply` runs.

Closes:

Relates:

### Description 📝

AWS Load Balancer Controller's mutating webhook sets `spec.loadBalancerClass` on the HAProxy `LoadBalancer` Services as soon as they are created while the controller is running, for example after migrating to HAProxy on an existing cluster. SD never declared that field. Once the webhook had set it out of band, the next `kapp` update touching these Services would fail with `spec.loadBalancerClass: ... may not change once set`, since the field is immutable once set. This triggered `kapp`'s `fallback-on-replace` strategy, which deletes and recreates the Service and its NLB.

This PR adds a `kapp` rebase rule so `spec.loadBalancerClass` always keeps whatever value is already on the live Service, instead of the module trying to change it.

### Breaking Changes 💔

None. 

### Tests performed 🧪

- [x] Reproduced the original failure end to end on a real EKS cluster: deployed HAProxy after the AWS Load Balancer Controller was already running, then upgraded the controller from v2.16.0 to v3.4.0 and the SD version. Confirmed the rebase rule prevents it. Service UID and NLB hostname stayed unchanged across the upgrade.

### Future work 🔧

None. 

> [!IMPORTANT]
>

- [x] CI is green


lukeskywlkr (approved):
LGTM! The rebase rule is a good idea to disarm it! 🚀

ralgozino (approved):

PR #594, opened 2026-09-07
docs(schemas): recommend secretbox for the encryption at rest

ralgozino:
### Summary 💡

The example of `spec.kubernetes.advanced.encryption.configuration` recommends a provider that upstream Kubernetes discourages, and it omits the `identity` provider. A cluster that follows the example cannot read the secrets that etcd holds from before the encryption was on.

### Description 📝

The example in the description field for etcd encryption now uses `secretbox` instead of `aescbc`, which the Kubernetes documentation discourages because CBC with PKCS#7 padding is open to padding oracle attacks.

Added the `identity` provider as last option in the example. A cluster that copied the example and turned the encryption on after creation could no longer read its own existing secrets.

Both `OnPremises` and `Immutable` carry the same description, thus both schemas change. The `EKSCluster` and `KFDDistribution` kinds have no such field.

Updated on-premises end-to-end tests and the configuration of the immutable schema test now use the same provider.

Relates:
- sighupio/furyctl#756 which adds `furyctl create secrets`. That command
writes this object with a new key, and it follows the example of this schema.

### Breaking Changes 💔

None. The schema accepts each provider that the API server accepts, and this PRchanges an example and three test fixtures.

### Tests performed 🧪

- [x] `jv` validates both schemas after the change
- [x] `jv` validates `tests/schemas/public/immutable-kfd-v1alpha2/002-ok.yaml` against the immutable schema, which is what `bats tests/e2e/immutable/schema.sh` runs
- [x] Both end-to-end configurations parse, hold a 32-byte key, and keep `identity` as the last provider

### Future work 🔧

None

~~- [ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)~~ N.A.
- [x] CI is green

### DIFF
--- a/docs/schemas/immutable-kfd-v1alpha2.md
+++ b/docs/schemas/immutable-kfd-v1alpha2.md
@@ -7168,21 +7168,27 @@ Maximum number of terminated Pods retained by the controller-manager before auto
 
 ### Description
 
-etcd's encryption at rest configuration. Must be a string with the EncryptionConfiguration object in YAML. Example:
+etcd's encryption at rest configuration. Must be a string with the EncryptionConfiguration object in YAML. Check the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/) for the providers that the API server accepts and for the properties of each one.
+
+Example:
 
 ```yaml
-
 apiVersion: apiserver.config.k8s.io/v1
 kind: EncryptionConfiguration
 resources:
   - resources:
-    - secrets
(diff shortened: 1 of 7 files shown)


stefanoghinelli (dismissed):
LGTM.
One non-blocking comment: we could align the [commented block](https://github.com/sighupio/distribution/blob/main/templates/config/onpremises-kfd-v1alpha2.yaml.tpl#L95-L104) in the OnPremises template with the schema

PR #596, opened 2026-09-14
fix(immutable): require arch on every node

### LINKED ISSUE #573: Immutable: the config generated by `furyctl create config` can't be applied
### Describe the bug

If you generate an Immutable config with `furyctl create config` and try to apply it, it fails while rendering the inventory. Two fields are missing from the generated file but the templates read them without a guard, and templates run with `missingkey=error`.

First one is `arch`. The generated config has it commented out with a note saying it's optional and defaults to `x86-64`:

```yaml
# arch: x86-64  # Optional: CPU architecture (x86-64|arm64). Default: x86-64.
```

The schema says the same thing — `arch` is not in `required` and it declares `"default": "x86-64"`. But nothing ever applies that default: furyctl passes the node straight from the YAML to the template, so a missing key stays missing, and `templates/kubernetes/immutable/hosts.yaml.tpl:16` reads `$n.arch` directly.

Also worth noting only 4 of the 11 nodes in the generated skeleton even have that commented line, so uncommenting the ones you can see isn't enough.

Second one is `pkiPath`. It isn't generated at all, and `hosts.yaml.tpl:202` reads `.spec.kubernetes.pkiPath`.

### To Reproduce

```bash
furyctl create config --kind Immutable --version v1.35.0 -c min.yaml
furyctl apply --config min.yaml --outdir . --workdir . --dry-run
```

```
template: hosts.yaml.tpl:16:93: executing "hosts.yaml.tpl" at <$n.arch>:
map has no entry for key "arch"
```

Add `arch: x86-64` to every node, run again, and you get the next one:

```
template: hosts.yaml.tpl:202:39: executing "hosts.yaml.tpl" at <.spec.kubernetes.pkiPath>:
map has no entry for key "pkiPath"
```

Add `pkiPath: ./pki` under `spec.kubernetes` and it renders fine.

### Expected behavior

`create config` should give you something that applies out of the box.

I think the cleanest fix is to make the templates honour the default that's already documented — something like `{{ $n.arch | default "x86-64" }}`, or guard the read. The other option is to stop calling these optional: emit `arch` on every node, generate `pkiPath`, and drop the `default` from the schema since nothing uses it. The first feels closer to what was intended.

### Software versions

- SIGHUP Distribution version: v1.35.0 (also reproduces on `main`, `af5d6181`)
- furyctl version: built from `main`, `332bc69a`

### Additional context

Found this while reviewing #570 — it's not related to that PR, it reproduces on `main` and the PR doesn't touch either template.

One more thing I noticed in the same file, tell me if you'd rather have it separately: `hosts.yaml.tpl:190-192` reads `$n.kernelParameters` inside `range $n := .spec.kubernetes.nodeGroups`, but `Spec.Kubernetes.NodeGroup` has `additionalProperties: false` and doesn't declare `kernelParameters`. So that block looks unreachable — validation would reject the field it needs.

ralgozino:
### Summary 💡

`furyctl create config --kind Immutable` produced a configuration file that `furyctl apply` could not use. Two node fields were missing; `pkiPath` was fixed in a later commit, `arch` was not. This makes `arch` required and writes it into the generated file, so the generated configuration applies without edits.

Closes: #573

Relates:
- sighupio/furyctl#760

### Description 📝

The Immutable schema declared `"default": "x86-64"` field for `spec.infrastructure.nodes[].arch`, but nothing applies the `default` field of a JSON schema.

The templates read the `arch` field without a guard, so a node without `arch` stopped the apply:

```
template: hosts.yaml.tpl:16:93: executing "hosts.yaml.tpl" at <$n.arch>: map has no entry for key "arch"
```

The field is read in four places, and a template guard reaches only two of them: the butane and iPXE templates render through a bare `text/template` with no function map at all. furyctl already backfilled the default by hand in one of the four (`infrastructure_bootstrap.go`), with a comment explaining that furyctl does not apply schema defaults — the same bug, discovered and worked around locally rather than fixed.

I decided to make the `arch` field required instead of making furyctl read the `deafult` field of the schema, we also have have similar logic with the `defaults` files but we can set a default there because the arch is defined inside an array that we don't know before hand.

Requiring the field removes the need to guard reading the arch on every template. It also fails better: a validation error that names the field, before anything runs, instead of a template error or `flatcar artifacts not found for architecture: ` with an empty value.

There is a safety argument too. `arch` selects which Flatcar artifacts and sysext packages are written to the machine, **so a wrong value (the default x86_64 instead of arm64, for example) produces a node that does not boot**. A value the user writes is safer than one furyctl assumes, above all for the mixed-architecture clusters this field exists to support.

For nodes with most configuration in common, a YAML anchor can help to avoid repeating configuration:

<details><summary>YAML Anchor example</summary>

```yaml
    nodes:
      - &node # YAML anchors to avoid repeating the same fields on all nodes
        hostname: cp1.example.com
        macAddress: "AA:BB:CC:DD:EE:FF"
        arch: x86-64
        network:
          ethernets:
            # Configure the bond0 to have a static address, gateway and nameserver
            eth0:
              dhcp4: true
        storage:
          installDisk: /dev/sda # main disk where to install Flatcar.
      - <<: *node
        hostname: cp2.example.com
        macAddress: "DD:EE:FF:AA:BB:CC"
```

</details>

The `default` keyword is gone — it was the only one in all of `schemas/public/`, and no code ever read it. The description also pointed at `examples/immutable-mixed-arch-example.yaml`, which does not exist in this repository; md-gen renders that description into the reference docs.

The second commit removes an unrelated block found while fixing this, mentioned in the issue: `hosts.yaml.tpl` set `sysctl_parameters` from `$n.kernelParameters` inside a range over `spec.kubernetes.nodeGroups`, note that it is taking the config from the `kubernetes` section, but for Immutable `kernelParameters` are defined in the `infrastructure` phase.

### Breaking Changes 💔

Not technically a breaking change because for 1.35.0 and 1.35.1 a cluster configuration that did not specify the node's architecture failed at apply time.

But worth noting that now `spec.infrastructure.nodes[].arch` becomes formally required. No working configuration is affected.

### Tests performed 🧪

- [x] Generated a configuration with `furyctl create config --kind Immutable --version v1.35.0` and applied it **unmodified** with `furyctl apply --dry-run`: it now renders the infrastructure and kubernetes phases and stops only at the distribution phase's kubectl call (no cluster, expected). On `main` the same command fails at `hosts.yaml.tpl:16`.
- [x] Diffed the rendered `kubernetes/hosts.yaml` against a pre-change render to confirm the `kernelParameters` removal changes no output — only the working-directory paths differ.
- [x] Confirmed the OnPremises `kernelParameters` block is untouched and still backed by `Spec.Kubernetes.Nodes.Node` in its schema.
- [x] Verified that an apply of a configuration file that does not specify a node's architecture now fails at validation time saying the field is missing.
- [x] Verified a correct configuration file still runs a full apply without errors.

### Future work 🔧

No CI step renders the Immutable kubernetes or infrastructure templates from a generated configuration, which is why this survived three months. `test-create-config-immutable` generates the file and stops, and `test:templates:regressions` runs `furyctl dump template`, which covers only the distribution phase — and within it only `manifests/`, `scripts/` and `_helpers/`, not `templates/distribution/terraform/`. Appending `furyctl apply --dry-run` to `test-create-config-immutable` would have caught this. There is no Immutable e2e pipeline yet.

- [x] CI is green


stefanoghinelli (dismissed):
LGTM (requires the docs generation)

